PR #11, opened 2017-04-12
Graft magma server

graft:
This branch creates a new Magma server, written in Rack. It has three endpoints: retrieve, query and update. Retrieve pulls records by model and record names, Update allows posts of revisions to records, and Query implements a new query API that does some sort of graph query.

There is also a Magma::Client; the magma gem has been slimmed down to just this client. There is now only a 'magma' gem, not a magma-dev-ucsf-immunoprofiler.org etc., since those models are only available to the server.

It also moves metrics to the backend from Timur.


Gemfile
```diff
@@ -1,7 +1,14 @@
...
-gem 'sinatra'
-
-gem 'magma-dev-ucsf-immunoprofiler'
-
+gem 'rack'
 gem 'pg'
+gem 'sequel'
+gem 'mini_magick'
+gem 'germ'
+gem 'net-http-persistent'
```
graft: This might not be necessary in the server Gemfile

config.ru
```diff
@@ -1,5 +1,13 @@
...
+require_relative 'lib/magma/server'
+require_relative 'lib/magma'
+require 'yaml'
+require 'logger'
 
-require File.expand_path '../magma-app.rb', __FILE__
+logger = Logger.new("log/error.log")
 
-run Sinatra::Application
+use Rack::CommonLogger, logger
```
graft: See how Metis etc. do rack logging

PR #14, opened 2017-05-10
clear cache on updates and new loader runs

graft:
Okay, this is an odd one.

The problem began when running a loader (either the TSV loader or the FlowJo loader, which loads data from a WSP file). Part of what the loader does is check if the record exists already; if it does, we need to update the database rather than insert a new row. In other cases it needs to dereference an identifier name to a foreign key id in another table - for example, a gene_exp record belongs to a rna_seq record - the loader entry might say "IPICRC032.rna.live", but what needs to be inserted into the database table is the foreign key for the corresponding rna_seq record.

Previously, I was simply using the ORM to check if the record exists or to get its id. This took a database query, which is slow, but I was not often inserting large numbers of records at once. Now I am; I need to be able to look up hundreds of thousands of record ids in sequence. I cannot do hundreds of thousands of database queries.

My solution is to memorize the identifier-id map for a model with a single query, and reused this cached list for checking my inserts (the method Magma::Model.identifier_id

However, there is a problem with this idea, which is that once I have inserted a new set of records with new identifiers, this map is out of date; that is, it will not recognize that my new records exist. (This resulted in an actual bug in WSP upload that Justin Lee encountered).

Therefore we need to invalidate the map (call the method Magma::Model.clear_cache) - the question is, when should we do it?

I have chosen to do it when a record is added to a model; this ensures that any records to be inserted/updated will be checked against a fresh map. However, I'm not sure this also takes care of the problem of the foreign key id lookup, which also uses the same map.

So the question is: do I invalidate the map in the right place? Is it possible to end up checking a stale map? Do I need to invalidate it elsewhere?


dabrau:
Not sure if we use postgres 9.5, but it has an "INSERT ON CONFLICT UPDATE". If the slowdown is the query I think you can batch it to a few queries vs hundreds of thousands. It might be worth trying out something like [https://gist.github.com/coin8086/a66c5f1a706b3981d1bdbe2cb7ff154d] to avoid the extra complexity of caching.

graft:
I'm not sure what postgres we use either - I remember contemplating using some upsert function in postgres (maybe this?) when I first wrote this thing. in lieu of upsert I do my own version of that active blob which splits the input and does a bulk update using a temporary table, which should work with any SQL db. It is pretty cheap, only two bulk insert operations (both done using Sequel's multi_insert) and an extra join/update for updates. However, you have to be able to split the input beforehand, thus the lookup.

Using upsert has some advantages - we use the native database constraint, it is guaranteed for each row, and it is probably faster than what I am doing. The main disadvantage I think is that it weds us to postgres, perhaps a particular version, but that might already be the case. There is also a bunch of other behavior we would have to replicate relating to updating foreign keys after an insert operation, but that is probably doable.

However, all of this would not spare us a query, because we would still have to "dereference" string identifiers to foreign key ids, so we have to make the query at least once.

Earlier I was maintaining the cache within the Magma::Loader object, so it would not last beyond one load operation - certainly not through several request cycles. All of this caching is just to support efficient loading, so that certainly makes sense. I moved it to Magma::Model because it seemed easier to reach for it there (there is also some other stuff being cached that I'm clearing up at the same time), but I think all of this trouble is just caused by that problem - the Magma::Model persists and carries the cache with it. Perhaps the correct solution here is just to move the caching back to the Loader object and have it go away when the loader is destroyed.

graft:
Okay, I made a new version of this which separates the validation into its own class, which does its own caching. Similarly the Loader class caches some queries to facilitate foreign key insertion. There might be more stuff like this in the future; this seems a reasonable paradigm - have the attribute spawn a class that has attribute-specific behavior, but can be thrown away when you are done with it.

PR #17, opened 2017-06-05
add innate stain to ipi models

graft:
This just adds the innate stain to ipi models - not really "magma", we need to split this code off into a magma-ipi repo.


graft:
Doh - there is also some stuff about reporting the parent of a model in the json template (and setting it in the model like we set 'identifier').

PR #21, opened 2017-09-12
Adding the germ gem repo

jasondcater:
Need to add the germ gem for 'bundler install'.


graft:
Should be a space between git: and '.

Err, i mean, merge...

PR #22, opened 2017-09-13
Graft fix update client

graft:
This mostly fixes an error with the Magma::Client#update method. It also adds some specs to test the client and makes a few crucial Gemfile edits.


jasondcater:
Please merge into etna-integration.

PR #23, opened 2017-09-19
Graft faster table

graft:
This adds a bunch of stuff!

1) The retrieve endpoint now works by generating a Question which it uses to retrieve data directly (avoiding the Sequel ORM).

2) There is now a 'Vector predicate' which is a valid child for a Record predicate. This allows you to map a list of values from a single Question (e.g., samples => [ sample_name, experiment_name, immune_count ]) (gets used by 1).

3) A 'filter' keyword on the retrieve endpoint, which subsets the Question using the Question Filter mechanism.

4) The retrieve endpoint has pagination, with support from the Question for computing page bounds.

5) The tsv format retrieve endpoint now returns an Enumerator, which allows streaming out of Rack. The TSV is loaded in pages (of 10,000) and streamed back to the user (this means no content-length is set). Note that if you test this, WebBrick does NOT correctly implement Rack streaming, so it will send the entire body at once and set a content-length.

5) The client gem takes a block which yields a response object, or merely returns a response object. The block response has NOT been read yet, allowing it to be read by chunks (with read_body).

6) The DocumentAttribute is renamed to FileAttribute


jasondcater:
This branch produces and error due to a missing configuration. In `lib/magma/storage.rb`

`    def get_url path
      if use_fog?
        @fog.get_object_url(
          @config[:directory],
          path,
          Time.now + @config[:expiration]*60,
          path_style: true
        )
      end
    end
`

`@config[:expiration]` needs to be defined. Please add a note here on what it should be.

graft:
Sure, it just sets the time in minutes that the get_url for an S3 link is valid before expiring. I set it to 1440, i.e., one day, which I think is an appropriate quantity.

jasondcater:
Seem like this is working ok. Please merge into etna-integration.

PR #24, opened 2017-09-30
Graft improve predicate interface

graft:
This cleans up the "predicate" classes that are used by the Query interface, to clarify the way they work. Mostly this consists of getting rid of switch statements and adding a "verb" interface that matches argument sequences in the query. The result is much cleaner classes that are easier to extend.

I have also added a few new verbs - ::any and ::count to the ModelPredicate and a bunch of new comparison/equality tests to the various value predicates.

There is also a reporting interface on the query endpoint (query('::predicates')) which returns the available predicates and the verb arguments they accept.

### DIFF
--- a/lib/magma/attributes/collection.rb
+++ b/lib/magma/attributes/collection.rb
@@ -23,7 +23,7 @@ class Magma
 
     def json_for record
       link = record[@name]
-      link ? link.map(&:last) : nil
+      link ? link.map(&:last).sort : nil
     end
 
     def txt_for record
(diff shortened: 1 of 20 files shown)


graft:
I added in two more issues that deal with collections; since I switched to using a LEFT OUTER JOIN to join tables, there are sometimes null rows. This was causing a [ null ] entry for empty collections; this should fix this.

Also the sorting for collections broke because of the retrieve endpoint; it used to use the ORM 'order' attribute; this is a less-good implementation, just sorting by identifier.

PR #26, opened 2017-10-04
fix sometime lack of attribute_names

graft:
This is the price I pay for weak test assertions - my update endpoint correctly changed the DB but returned a 500 instead of a 200 due to some stupid error. I added better assertions and fixed the error.

### DIFF
--- a/lib/magma/payload.rb
+++ b/lib/magma/payload.rb
@@ -101,8 +101,9 @@ class Magma
       def json_document record
         # A JSON version of this record (actually a hash). Each attribute
         # reports in its own fashion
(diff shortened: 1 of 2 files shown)


lib/magma/payload.rb
```diff
@@ -101,8 +101,9 @@ def to_hash
       def json_document record
         # A JSON version of this record (actually a hash). Each attribute
         # reports in its own fashion
+        names = attribute_names || @model.attributes.keys
```
dabrau: @attribute_names?

graft:
I guess that is using the attr_reader :attribute_names, I suppose @attribute_names is clearer.

PR #27, opened 2017-10-05
add uploads/ to s3 storage path

graft:
Following the use of the Question interface to get records, we now use fog's get_url method to generate an s3 url (instead of letting carrierwave do it through the Sequel model) - however, carrierwave seems to put files into an 'uploads/' directory in s3. We need to format our path correspondingly.


dabrau:
looks good to me.

PR #30, opened 2017-10-12
Graft fix linking and validation

graft:
This does two main things:

1) Adds Magma#get_project, Magma#magma_projects, which make use of a new Magma::Project class which collects all of its own models and so on.

2) Adds in a bunch of validations when the model loads:
  - There must be a Project model for each project
  - Model names cannot be certain reserved words
  - Models must have reciprocal links - if A links to B, B must link to A somehow.
  - Models cannot be orphans - they must link to another model

There is a problem with the orphan model test, which is that you can have two or more orphans clutching each other and still disconnected from the Project model, really we should validate that there is a path to Project somehow.

### DIFF
--- a/lib/magma.rb
+++ b/lib/magma.rb
@@ -47,7 +47,7 @@ class Magma
     (ENV['MAGMA_ENV'] || :development).to_sym
   end
 
-  def load_models(check_tables = true)
+  def load_models(validate = true)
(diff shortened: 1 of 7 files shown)


jasondcater:
This has been working well for me. Go ahead and merge it.

PR #31, opened 2017-11-07
fix retrieve filtering on numbers

graft:
Pretty simple bug-fix, just deals with lack of '.to_f' on a string input value.


graft:
I added another bug fix which is a pretty big deal, putting constraints from the "Vector" predicate (i.e., one query that maps many values) into the SQL query - without this the query becomes huge and crashes magma.

PR #32, opened 2017-11-17
Graft use etna server

graft:
This just moves Magma over to Etna::Server, etc., depending on the Etna gem instead of its own classes - mostly this just involves deleting code and renaming classes.


jasondcater:
The magma client, in this branch, doesn't work with the master branch of Timur. We either need to get it working with the master branch of Timur or wait until Timur starts using the etna application gem.

jasondcater:
This works. Merge to master and I'll push this to staging.

PR #68, opened 2018-02-13
Da tsv large tables

dabrau:
- modified the question class to query once for the page bounds when iterating through pages
- created a module to share some code to write a tsv
- created a class to unload a model as a tsv


graft:
I think tsv writer should be a class rather than a module (instantiated by e.g. Retrieve), I find it clearer to see where methods are coming from that way.

spec/tsv_writer_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        @lines.push(line)
+      end
+    end
+
+    file = MockFile.new
+    payload = Magma::Payload.new
+    TSVWriter.new(model, retrieval, payload).write_tsv(file)
+
+    header = "project\tname\tmonster\tnumber\tcompleted\tyear\n"
+    expect(file.lines[0]).to eq(header)
```
graft: I would build this using retrieval.attribute_names or model.attributes.keys or something so it isn't hard-coded.

spec/tsv_writer_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      def initialize
+        @lines = []
+      end
+
+      def << line
+        @lines.push(line)
+      end
+    end
+
+    file = MockFile.new
```
graft: Might want to use StringIO.new here, which will give a text blob that might be more like the actual file.

lib/magma/commands.rb
```diff
@@ -127,4 +127,26 @@ def setup(config)
       Magma.instance.load_models
     end
   end
+
+  class Unload < Etna::Command
+    usage 'Run Unloader to dump the dataset of a model into a gzipped tsv.
+      *args - arg[0]: project_name, arg[1]: model_name, args[2]: gzip?, arg[3]: file'
```
graft: The usage i've been favoring is like:
`<project_name> <model_name> <gzip?> <file> # Dump the dataset of the model into a gzipped tsv`

lib/magma/commands.rb
```diff
@@ -127,4 +127,26 @@ def setup(config)
...
+
+    def execute(*args)
+      begin
```
graft: this could be execute(project_name, model_name, gzip?, directory), so you don't have to use args[0] etc. below

lib/magma/unloader.rb
```diff
@@ -0,0 +1,30 @@
+require_relative './tsv_writer'
+require_relative './retrieval'
+
+class ModelUnloader
```
graft: This should probably be Magma::Unloader since the file is named that way

graft:
I would favor getting rid of the gzip stuff, since it's easily done with gzip -c in a shell script and is probably faster that way; in fact I'd get rid of all file writing entirely and just write to stdout. It is the most unix-like, the most flexible, and makes the least assumptions about how this command is used (for example, you can feed it to grep that way). You can still fake this by passing in /dev/stdout as the command parameter but I think it works better that way. Probably means you can get rid of your ModelUnloader command.

lib/magma/unloader.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def tsv_dump(options = {})
+    gzip = options[:gzip]
+    file_dir = options[:file_dir]
+
+    if !file_dir
+      time = Time.now
+      # default file_dir format e.g. "/var/tmp/project_model_2018-2-13_dump.tsv"
+      file_dir = "/var/tmp/#{@model.project_name.to_s}_#{@model.model_name.to_s}_#{time.year.to_s}-#{time.month.to_s}-#{time.day.to_s}_dump.tsv"
```
graft: We shouldn't hard-code any file locations since we can't assume /var/tmp exists or is writeable

spec/tsv_writer_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+        nil,
+        1,
+        2
+    )
+
+    file = StringIO.new
+    Magma::TSVWriter.new(model, retrieval, payload).write_tsv{ |lines| file.write lines }
+
+    lines = file.string.split("\n")
+    expect(lines[0]).to eq(payload.tsv_header.sub("\n", ''))
```
graft: This should probably be something like expect(lines[0]).to eq(model.attributes.keys.join('\t'))

spec/tsv_writer_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(lines.size).to eq(5)
```
graft: This could test lines =~ labors.map(&:name) or something like that.

graft:
Other than that shmeeny spec change this looks good to me.

PR #72, opened 2018-02-26
Graft fix retrieve filtering

graft:
Couple of small fixes as I try to make this request endpoint available to outsiders, mostly a bug in parsing date-times.


dabrau:
Looks good. :shipit: 📦

Issue #65, opened 2018-02-02
Complex queries do not time out.


graft:
Because of large holes in the query design, it is easy for users to construct queries that result in very large tables being joined together, perhaps in triplicate. This is very bad and must be solved.

However, until that problem is solved, users can easily lock up Magma with long-running queries just by hitting reload on their favorite manifest a few times. This locks up Timur, which can't do anything without Magma data.

To prevent this disastrous situation from persisting, we can use the postgres variable statement_timeout; probably by adding this setting to the database role for magma. This way at least those queries will die and restore normal function within whatever statement_timeout is - maybe one minute?

PR #73, opened 2018-03-01
query time out

dabrau:
related to this issue #65 

- Manage the query timeouts within the question, so that database roles don't have to be changed and timeouts can be managed for read queries

- query is done inside a transaction with a local statement_timeout, only works for postgres

- timeout set to 5 minutes in the query controller

### DIFF
--- a/lib/magma/server/query.rb
+++ b/lib/magma/server/query.rb
@@ -6,10 +6,10 @@ class QueryController < Magma::Controller
     return failure(401, errors: [ 'You are unauthorized' ]) unless @user && @user.can_view?(@project_name)
     begin
       return success(Magma::Predicate.to_json, 'application/json') if @params[:query] == '::predicates'
-      question = Magma::Question.new(@project_name, @params[:query])
+      question = Magma::Question.new(@project_name, @params[:query], :timeout => 300000)
       return_data = {answer: question.answer, type: question.type}
       return success(return_data.to_json, 'application/json')
-    rescue ArgumentError => e
+    rescue StandardError => e
       puts e.backtrace
       return failure(422, errors: [ e.message ])
     end
--- a/lib/magma/server/query.rb
+++ b/lib/magma/server/query.rb
@@ -6,12 +6,16 @@ class QueryController < Magma::Controller
     return failure(401, errors: [ 'You are unauthorized' ]) unless @user && @user.can_view?(@project_name)
     begin
       return success(Magma::Predicate.to_json, 'application/json') if @params[:query] == '::predicates'
-      question = Magma::Question.new(@project_name, @params[:query], :timeout => 300000)
+      question = Magma::Question.new(@project_name, @params[:query],
+                                     :timeout => Magma.instance.config(:query_timeout))
       return_data = {answer: question.answer, type: question.type}
       return success(return_data.to_json, 'application/json')
-    rescue StandardError => e
+    rescue ArgumentError => e
       puts e.backtrace
       return failure(422, errors: [ e.message ])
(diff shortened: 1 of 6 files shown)


lib/magma/server/query.rb
```diff
@@ -6,10 +6,10 @@ def action
     return failure(401, errors: [ 'You are unauthorized' ]) unless @user && @user.can_view?(@project_name)
     begin
       return success(Magma::Predicate.to_json, 'application/json') if @params[:query] == '::predicates'
-      question = Magma::Question.new(@project_name, @params[:query])
+      question = Magma::Question.new(@project_name, @params[:query], :timeout => 300000)
```
graft: Can we put this in a config variable like :query_timeout?

lib/magma/server/query.rb
```diff
@@ -6,10 +6,10 @@ def action
...
       return_data = {answer: question.answer, type: question.type}
       return success(return_data.to_json, 'application/json')
-    rescue ArgumentError => e
+    rescue StandardError => e
```
graft: Does the timeout generate a more specific error? I would probably catch that rather than catching StandardError, since not every StandardError implies a client error. In this case I'm not sure a timeout is a 422, it should probably be something like a 500...

graft:
Hmm, any way to test this in a spec? Timecop won't help here... maybe you could set the timeout to be very small?

lib/magma/server/query.rb
dabrau: How do you feel about catching the DatabaseError and sending a 500? I don't think a 504 is appropriate  because other things can raise a DatabaseError.

dabrau:
Was thinking about adding an execute_query method in the Magma singleton. That way it can be tested by passing in a sql query 'pg_sleep' and abstract the implementation from 'question'. Thoughts?

lib/magma/server/query.rb
graft: Yeah in general I feel like 500 is all they need to know. Definitely not a client error, they just submitted a well-formulated query. It's our fault we can't fulfill it, so probably a 501 'Not Implemented'

graft:
I wouldn't put this stuff in the Magma singleton, the point of the Question was to isolate query logic there, so I think it should remain there. I think it would also be hard to stub out this low-level method, you're already reaching past Magma into Sequel with @db.run and @db.transaction. What you want is to catch all of the output Magma/Sequel makes to postgres. I bet someone has written a pg_stub type gem - or perhaps you could just catch at the TCP level with something like WebMocks?

But if so I think the spec should be:
test the /query endpoint with a simple query
Stub the postgres connection somehow and catch the SQL commands being run
Verify that the query contains SET LOCAL statement_timeout = whatevs

dabrau:
To me the point of the Question looks like it transforms the DSL to a query and should delegate the execution to something like QueryExecutor. I think it makes sense to be tested without the context of a Question.

- test query execution method, send the correct SQL and that it throws an error
- test question.answer makes a call to the execution method
- test the endpoint has the correct response when question.answer throws a Database Error

graft:
I agree the Question should delegate to some other class, just not the Magma singleton where the "context" is too broad. In this case your first test is good. I think the second test should also just be an endpoint test for correct response without hitting the timeout, I think it is bad to enshrine Question#answer as important in a spec. It seems like all of these will involve trapping the SQL

dabrau:
I am limiting the testing to ensure the timeout works. The second test you describe sounds like a test that is covered by the query_spec. Trapping SQL only makes sense for testing how the query gets executed.

PR #76, opened 2018-03-08
support logging in db again

graft:
The Sequel query-logging seems to have been axed in the etna-server update, adding it back in.


dabrau:
:shipit:

PR #77, opened 2018-03-12
fix retrieve endpoint to re-allow grabbing identifiers (only) for all records/models

graft:
A victim of poor test coverage, a previous change, meant to prevent 'grabbing the entire dataset' also prevented 'grabbing just the identifiers for the entire dataset'. I fixed, added in a test to cover the specific case, and made some stylish improvements to the spec.

### DIFF
--- a/spec/retrieve_spec.rb
+++ b/spec/retrieve_spec.rb
@@ -43,16 +43,16 @@ describe Magma::RetrieveController do
       project_name: 'labors',
       model_name: 'all',
       record_names: [],
-      attribute_names: "all"
+      attribute_names: 'all'
     )
     expect(last_response).to(be_ok)
   end
 
-  it "complains if there are record names for all models" do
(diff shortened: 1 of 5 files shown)


spec/retrieve_spec.rb
```diff
@@ -64,12 +64,28 @@ def retrieve(post, user_type=:viewer)
...
+      model_name: 'all',
+      record_names: 'all',
+      attribute_names: 'identifier'
+    )
+    expect(last_response.status).to eq(200)
+    json = json_body(last_response.body)
+
+    expect(json[:models][:labor][:documents].keys).to eq(labors.map(&:name).map(&:to_sym))
+    expect(json[:models][:monster][:documents].keys).to eq(monsters.map(&:name).map(&:to_sym))
+  end
```
dabrau: You might want to test what models return an identifier and what doesn't.

PR #81, opened 2018-05-06
adds etna::cross_origin layer so magma can serve requests to browser clients

graft:
This lets magma use the new etna cross_origin rack layer, which allows it to respond correctly to CORS requests from browsers (so, e.g., timur.ucsf.edu can talk directly to magma.ucsf.edu via the browser).


jasondcater:
Looks good. Please merge.

PR #82, opened 2018-05-06
allow each project repo to track its own gems and bundle with magma's Gemfile

graft:
Since projects have loader code, they will inevitably accumulate external dependencies. Rather than adding them to the base magma Gemfile, this simply loads a Gemfile from each projects/whatever/ directory, so each project can define its own dependencies. Note this assumes that the project_path is projects/*, which may not be true depending on the individual config.yml.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,3 +21,7 @@ group :test do
   gem 'pry'
   gem 'timecop'
 end
+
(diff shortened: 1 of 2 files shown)


jasondcater:
Looks good. Please merge.

PR #90, opened 2018-08-30
Graft fix migrations

graft:
This pull request cleans up a few migration-related issues regarding Sequel table aliases, deprecation warnings, migration formatting, and so on. This also adds a lot of tests to cover migration generation.

Magma is still pegged to run with Sequel 4.49 and does not seem to run with Sequel 5, which should be fixed at some point, but that doesn't happen here. Gemsets containing Sequel 5 might require using 'bundle exec' to run rspec tests.

### DIFF
--- a/lib/magma/migration.rb
+++ b/lib/magma/migration.rb
@@ -40,6 +40,24 @@ class Magma
       end.join("\n").chomp
     end
 
+    def attribute_migration(att)
+      case att
+      when Magma::ForeignKeyAttribute
+        [
+          foreign_key_entry(att.column_name, att.link_model),
+          index_entry(att.column_name)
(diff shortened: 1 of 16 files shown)


lib/magma/migration.rb
```diff
@@ -1,8 +1,11 @@
 class Magma
   class Migration
     class << self
+      def table_name(model)
+        "Sequel[:#{model.project_name}][:#{model.implicit_table_name}]"
+      end
       def create(model)
-        puts model.table_name
+        #puts table_name(model)
```
dabrau: uncomment or delete

dabrau:
Looks good. ship it.

Issue #36, opened 2017-12-06
Add PHI features to Magma attributes


graft:
Any attribute in Magma can be flagged as 'phi'. If this is the case, Magma will:

1) Return the data if the user has permission 'phi' for the project. 

2) Return a transformed version of the data if the user lacks permission 'phi' for the project.

The transformation should be consistent - the easiest way is to compute it using a hash of an associated identifier (e.g. a patient identifier). The 'phi' attribute should therefore both point to the identifier used for anonymizing and specify an anonymization strategy.

graft:
Importantly, the /retrieve and /query endpoints must censor the data appropriately.  If the attribute is flagged 'restricted' (what Janus calls this flag), /query and /retrieve should both refuse to show that data to a user without the restricted permission.

How should the censoring operate?

1. Should you remove restricted attributes from the JSON template, or leave the template intact describing what the user can't see?

2. Should you remove all restricted data from the JSON document describing a record, or merely replace it with a censored value? If the latter, how is the censored value computed?

jasondcater:
From previous conversations we should remove/censor information while leaving the metadata intact. i.e. show that the field is present but remove the data.

graft:
This is being implemented using the 'restricted' project permission on Janus. As Magma treats it, a record can have a 'restricted' attribute, or an attribute can be flagged as 'restricted' in the model. If the former is 'true', only project members with restricted permission can see anything from that record. If the latter is 'true', only project members can see data from that attribute in any record.

graft:
How should censoring be done? All three endpoints, /update, /retrieve and /query, can yield potentially restricted data. Each should be censored to ensure that data does not leak out.

Data is accessed from these endpoints in a variety of ways - for example, the retrieve endpoint can return tsv or json-formatted data, with slightly different queries posted. At the end of the day, however, all three endpoints use a Magma::Question, which takes a query in Magma's query language, builds a corresponding SQL query, and executes that. It would seem logical, then, to perform the censoring here, rather than separately in each endpoint, to ensure that data cannot escape in uncensored form.

In order to accomplish this the Question needs to be aware of the restricted permissions, which it currently is not. Each endpoint has to pass the restriction (@user.can_see_restricted?(@project_name)) in to the Question. The /retrieve and /update endpoints both do this via the Magma::Retrieval object, which must be modified accordingly.

Alternative strategies:

1) Censor data as it comes out. This might be difficult as the results of a query would have to be re-parsed to understand the context.

[...]

PR #92, opened 2018-09-11
Graft add restricted

graft:
This PR implements #36, allowing data restrictions to records which censors certain pieces of data depending on user permissions. This relies on the janus 'restricted' permission, which can be set for each user for each project, independent of their role (so you can be a restricted admin, or an unrestricted viewer).

A note on nomenclature - I am using 'restricted user' to mean "cannot access restricted data", and 'unrestricted user' to mean "can access restricted data". This nomenclature might be annoying or confusing - however, I think it's pretty clear how it actually applies to the data (restricted data is censored, non-restricted data is visible). 

There are two ways to restrict data. The first is adding the 'restricted' attribute, defined on a model as so:

    class MyModel < Magma::Model
      restricted
    end

This adds a boolean attribute named 'restricted' to the model. This attribute may be queried as normal. However, restricted users cannot retrieve or modify ANY data from a record where 'restricted' is true.

The second way is to add 'restricted' to any attribute, defined like so:

    class MyModel < Magma::Model
      attribute :my_attribute, restricted: true
    end

This prevents any restricted user from retrieving data in this attribute from ANY record (whether or not it is 'restricted' as in the first case).

If you think of the model/records as a table, the first method censors rows (records), the second censors columns (attributes).

The way this is implemented is by passing the instruction to restrict into Magma::Question and Magma::Revision, which either filter records (first method) or raise an error (second method) if the data should not be visible.


lib/magma/retrieval.rb
```diff
@@ -10,23 +10,25 @@ def initialize(model, record_names, attribute_names, opts={})
...
-      @collapse_tables = opts[:collapse_tables]
-      @page_size = opts[:page_size]
-      @page = opts[:page]
     end
 
     def attributes
       @attributes ||= begin
         attributes = @model.attributes.values.select do |att|
-          !att.is_a?(Magma::TableAttribute) && requested?(att)
+          !att.is_a?(Magma::TableAttribute) && requested?(att) && !restricted?(att)
```
dabrau: This could be confusing. Attributes that don't exist and attributes that are restricted are handled the same. You might want to return some meta data to the users to specify attributes that don't but are requested, attributes you are restricted from, and whether the query is run as a restricted user.

graft: The user can ascertain what attributes are available, restricted and not from the json template, which is maybe adequate documentation. I'm not sure about adding rejected arguments to the return value - it might be useful to raise an argument error if bad or restricted attributes are requested instead

PR #93, opened 2018-10-05
fix pagination issue in retrieve

graft:
At some point (due to a rather poorly-written interface in the function RetrieveController::retrieve_model) a regression was introduced when using the option collapse_tables (which removes table attributes from the results, used, e.g., by the Search page) that caused all pages to return instead of a specific page. This PR fixes that issue and adds a test to cover this case.


dabrau:
:shipit:

PR #94, opened 2018-10-05
Update Ruby to 2.5.1

graft:
Updating Magma to use Ruby 2.5.1. The only issue I found was with the esoteric entity 'ObjectSpace' which I was using to find derived classes.


eleanormark:
Magma ran properly with ruby version  update to 2.5.1.

eleanormark:
Magma ran properly with Ruby 2.5.1 upgrade.

PR #95, opened 2018-10-20
adjust model ordering in project migration

graft:
This is a fix to adjust the way tables are output when planning a migration.

The existing plan command outputs tables in some unspecified order - this can result in an initial migration which does not work, because a create_table may contain foreign keys to tables that do not exist yet.

This PR attempts to fix that by outputting tables in a breadth-first order based on parent models, to attempt to ensure that there are no foreign key conflicts. This doesn't prevent all situations from causing trouble (e.g. circular links between models), but it should do most of the time.

### DIFF
--- a/lib/magma/project.rb
+++ b/lib/magma/project.rb
@@ -21,8 +21,15 @@ class Magma
       ]
     end
 
+    def ordered_models(model)
+      link_models = model.attributes.values.select do |att|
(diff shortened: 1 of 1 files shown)


graft:
A potential future adjustment: output all create_tables first, without foreign key columns. Subsequently, use alter_table to add the foreign keys.

graft:
Only touches the plan command and seems to work as intended, merging.

Issue #100, opened 2019-01-28
/query excludes rows with empty data


graft:
Many queries take the form of returning a "Vector" predicate (which should probably be renamed to "Table" predicate or something) - the intent is to return an array of results for each row in the query. E.g.

    [ model, ...filters, '::all', [ result1, result2, result3 ] ]

where result1, etc., map the returned model items to output values (e.g. [ model_column ] being the simplest possible mapping).

Often these results will map columns from same table several times; for example, if I have a 'sample' model, I might want to map several 'sample.gene_exp.count' values, one for each value of 'gene_exp.name', i.e., for each sample I might want to know the counts for GENE1, GENE2, and GENE3. The resulting query might look like this:

    [ 'sample', ['sample_name', '::matches', 'BOB' ], '::all',
      [
        [ 'gene_exp', [ 'name', '::equals', 'GENE1' ], '::first', 'count' ],
        [ 'gene_exp', [ 'name', '::equals', 'GENE2' ], '::first', 'count' ],
        [ 'gene_exp', [ 'name', '::equals', 'GENE3' ], '::first', 'count' ]
      ]
    ]
This query produces a SQL statement which uses LEFT OUTER JOIN to join the sample table to the gene_exp table three times (with three different aliases for gene_exp). Currently the statement will join on the foreign key (i.e. `ON sample.id = gene_exp1.sample_id`); subsequently the joined table will be filtered using WHERE clauses to reduce it to the relevant rows, i.e. `WHERE gene_exp1.name = 'GENE1' AND gene_exp2.name = 'GENE2' AND gene_exp3.name = 'GENE3'`

In the situation where the gene_exp table lacks a row containing name='GENE1' but has a row containing name='GENE2' and 'GENE3', the WHERE condition will fail (there is no row in the join table satisfying all three AND clauses), and there will be no data returned. This is not what we want and defeats the whole purpose of using a LEFT OUTER JOIN, which is that we want there to be NULL rows when data is missing.

A solution is to move the WHERE conditions into the ON clause; this will mean the table will return rows even with missing data.

graft:
My initial attempt produced some problems; while moving the WHERE conditions into the ON clause solved the initial problem it produced a bunch of other failing tests. The problem is that now we are unable to eliminate rows on filtering, because all of the filters are now being LEFT OUTER JOINed - clauses that are meant to reduce rows in the table instead retain the row with a NULL value.

[...]

PR #101, opened 2019-03-21
Graft fix query

graft:
This PR solves a bug in the /query API that removed some items from the response incorrectly, due to outer-joining tables on foreign key criterion (correctly producing rows with NULL values) but then applying filters to these rows in WHERE clauses (incorrectly remove rows with some NULL values). See discussion of the problem in #100 - the resolution was to move filtering criterion into the JOIN ON clause for mapped columns, but to keep those criterion in WHERE clauses for filters on the table.

The final result was incompatible with the current plot code, so this has been waiting on the completion of the timur branch `em-d3-v5`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,10 +6,10 @@ gem 'etna'
 gem 'pg'
 gem 'sequel', '4.49.0'
 gem 'mini_magick'
+gem 'fog-aws'
 gem 'carrierwave-sequel'
 gem 'carrierwave'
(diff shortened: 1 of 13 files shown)


graft:
There are also some changes regarding the fog gem, which was previously loading a bunch of requirements (like fog-google) instead of just fog-aws. There are also some Etna logging changes (including hiding a database error instead of broadcasting it, wtf?).

Issue #113, opened 2020-03-25
Hide type in attributes/split base Attribute


graft:
Currently the Magma::Model.attribute method accepts a :type option. This allows you to set the database column type (via the Sequel ORM, which will set the postgres data type for you). This allows for some interesting non-desirable behavior in the model space, if an attribute can be set with a column type that Magma does not directly support.

Especially in the era of #104 where we will not be programming models (and thus allowing all sorts of non-standard shenanigans like this) but instantiating from data, we should instead have only one exposed set of types, that is, magma types.

The current 'type' columns in use are DateTime, Float, String and Integer.  There is actually quite a lot of code in Magma and Timur splitting on Magma::Attribute#type as a result of this division, which forks the basic Attribute into these four subtypes.

The change we wish to make is:

1. Make Magma::Model.attribute private
2. Add new attribute classes (Magma::DateTimeAttribute, Magma::FloatAttribute, Magma::StringAttribute, Magma::IntegerAttribute) and their corresponding Magma::Model methods (e.g. Magma::Model.integer). These methods should set type, but not expose it as an option.
3. Update model specs to use 'integer :att_name' instead of 'attribute :att_name, type: Integer'
4. Fix ColumnPredicate, etc. to use new attribute classes rather than Magma::Attribute.

graft:
Completed by #120

PR #118, opened 2020-04-07
Change attribute definitions

notmarkmiranda:
#113 

This will start work on the linked ticket. It will make the `#attribute` method private within the `model` class and allow for attribute creation by using `data_type :attribute_name` where the allowed data_types are `string`, `integer`, `boolean`, and `date_time`.

### DIFF
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -136,6 +136,5 @@ require_relative 'attributes/string_attribute'
 require_relative 'attributes/integer_attribute'
 require_relative 'attributes/boolean_attribute'
 require_relative 'attributes/date_time_attribute'
-require_relative 'attributes/json_attribute'
 require_relative 'attributes/float_attribute'
 
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -9,9 +9,8 @@ class Magma
 :format_hint, :loader, :link_model, :restricted ]
       end
 
-      def set_attribute(name, model, options, caller)
-        klass = "Magma::#{caller.to_s.capitalize}_attribute".camelcase.constantize
-        klass.new(name, model, options)
+      def set_attribute(name, model, options, attribute_class)
+        attribute_class.new(name, model, options)
       end
     end
 
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -3,11 +3,15 @@ class Magma
     DISPLAY_ONLY = [:child, :collection]
     attr_reader :name, :type, :desc, :loader, :match, :format_hint, :unique, :index, :restricted
 
-
     class << self
       def options
         [:type, :desc, :display_name, :hide, :readonly, :unique, :index, :match,
 :format_hint, :loader, :link_model, :restricted ]
+      end
+
+      def set_attribute(name, model, options, caller)
+        klass = "Magma::#{caller.to_s.capitalize}_attribute".camelcase.constantize
+        klass.new(name, model, options)
       end
     end
 
@@ -129,3 +133,10 @@ require_relative 'attributes/match'
 require_relative 'attributes/image'
 require_relative 'attributes/table'
 require_relative 'attributes/matrix'
+require_relative 'attributes/string_attribute'
+require_relative 'attributes/integer_attribute'
+require_relative 'attributes/boolean_attribute'
+require_relative 'attributes/date_time_attribute'
+require_relative 'attributes/json_attribute'
+require_relative 'attributes/float_attribute'
+
--- a/lib/magma/attributes/file.rb
+++ b/lib/magma/attributes/file.rb
@@ -1,8 +1,9 @@
 class Magma
   class FileAttribute < Attribute
     def initialize(name, model, opts)
-      super
       @type = String
+      Magma.instance.storage.setup_uploader(model, name, :file) 
+      super
     end
 
     def update(record, new_value)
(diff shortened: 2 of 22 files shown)


lib/magma/model.rb
```diff
@@ -4,8 +4,25 @@
...
+      def integer(attribute_name, opts = {})
+        attribute(attribute_name, opts.merge(type: Integer))
+      end
+
+      def boolean(attribute_name, opts = {})
+        attribute(attribute_name, opts.merge(type: Boolean))
+      end
+
+      def date_time(attribute_name, opts = {})
+        attributes(attribute_name, opts.merge(type: DateTime))
```
graft, line 23: should be attribute()

lib/magma/model.rb
```diff
@@ -4,8 +4,25 @@
...
+      def string(attribute_name, opts = {})
+        attribute(attribute_name, opts.merge(type: String))
+      end
+
+      def integer(attribute_name, opts = {})
+        attribute(attribute_name, opts.merge(type: Integer))
+      end
+
+      def boolean(attribute_name, opts = {})
+        attribute(attribute_name, opts.merge(type: Boolean))
```
graft, line 19: There's no Boolean class, so I think this must be TrueClass in order to work.

lib/magma/model.rb
```diff
@@ -238,8 +249,16 @@ def inherited(magma_model)
         )
 
         super
-        magma_model.attribute(:created_at, type: DateTime, hide: true)
-        magma_model.attribute(:updated_at, type: DateTime, hide: true)
+        magma_model.send(:attribute, :created_at, type: DateTime, hide: true)
```
graft, line 252: could these instead use the new interface, e.g. magma_model.date_time(:created_at, hide:true)?

lib/magma/model.rb
```diff
@@ -238,8 +249,16 @@ def inherited(magma_model)
...
+        magma_model.send(:attribute, :updated_at, type: DateTime, hide: true)
+      end
+
+      private
+
+      # basic attribute, holds any pg data type
+      def attribute(attr_name, opts = {})
+        klass = opts.delete(:attribute_class) || Magma::Attribute
```
graft, line 260: might be better to start requiring attribute_class as a method argument rather than passing it in opts since we want to deprecate using Magma::Attribute as a base class (i.e., attribute_class will be required).

notmarkmiranda:
I think because you used `class << self` the private actually works. I threw a pry in one of the tests and attempted to call `Magma::Model.attribute` and get a `NoMethodError`.

image

graft:
While that works if you stick a pry in class Labors::Aspect e.g. you are
inside the class and can access #attribute.

lib/magma/attribute.rb
```diff
@@ -3,12 +3,16 @@ class Attribute
...
 
-
     class << self
       def options
         [:type, :desc, :display_name, :hide, :readonly, :unique, :index, :match,
 :format_hint, :loader, :link_model, :restricted ]
       end
+
+      def set_attribute(name, model, options, caller)
+        klass = "Magma::#{caller.to_s.capitalize}_attribute".camelcase.constantize
```
graft, line 13: Could probably move this stuff to the #define_method call below and avoid the caller argument.

graft:
A few observations:

1) Magma::Model._attribute remains and continues to be used by .collection, .matrix, etc. - since _attribute does the same work as set_attribute this interface confusion ought to be resolved. This division is also reflected in the list of attribute classes used to define #integer, etc. in Magma::Model, which includes the previously classless attributes but excludes the existing attribute classes (CollectionAttribute, MatrixAttribute, etc.).

2) There is a 'json' attribute that i don't think should exist. While in the future we might expose a generic json attribute, for the moment json columns are used only in specific ways (to hold the data for the match type or the matrix type) and should not be exposed as an attribute class.

3) The new use of IntegerAttribute etc. is not reflected in case statements like the one in #attribute_child in lib/magma/query/predicate/record.rb, which switches on attribute.type. The code still works because IntegerAttribute inherits from Magma::Attribute and exposes #type, so the method continues to function. Probably, however, we should hide Magma::Attribute#type since it does not need to be exposed and resolve this case statement to use the attribute class rather than the type.

lib/magma/attribute.rb
notmarkmiranda: 👍🏼

notmarkmiranda:
`Labors::Prize` and `Labors::Olympians` tests have `json: ` attributes. If I remove that, what type should that be stored as?

lib/magma/attributes/file.rb
```diff
@@ -1,8 +1,9 @@
 class Magma
   class FileAttribute < Attribute
     def initialize(name, model, opts)
-      super
       @type = String
+      Magma.instance.storage.setup_uploader(model, name, :file)
```
graft, line 5: ImageAttribute inherits from this method, but does not replace this setup_uploader call with one using the :image argument. I think the uploading code is poorly-tested, so this might have slipped under the radar.

notmarkmiranda: I'm not sure how to fix this? I can add an initialize method to ImageAttribute, but there's nothing stop this from being run in the ImageAttribute#new and then again in FileAttribute#new when super is called?

graft:
Could either infer the type from self.class.name or put it all in a setup
method called in the initializer and overriden by imageattribute.

PR #122, opened 2020-04-20
Remove attribute usage in example proejct models

alimi:
Per https://github.com/mountetna/magma/pull/118, attribute is no longer available in `Magma::Model`s.


graft:
:shipit:

Issue #99, opened 2019-01-25
Hot editing of models


graft:
Magma models are fairly rigid data structures at this point - they are defined in a ruby class, which must be updated by creating a new branch for the project repo, editing the appropriate model file and updating an attribute, creating a pull request, deploying it to staging, etc. This is a very slow process.

Changes that add or remove attributes or introduce new models require writing a database migration. These changes might be more substantial and it might be important to do them by hand. Still, we might wish for tooling to make this process easy for us.

However, the other case, of changes to the models that do NOT require a migration, but merely change the description, match, etc., of an attribute do NOT need to live in the model. While the model might represent a sensible default, we can make these attributes user-accessible by moving them into a database.

Exposing these (especially the attribute.match, which performs validation) allows users (or at least superusers) to more directly shape usage of their models, for example by directly inputting and updating their preferred identifier pattern; it also reduces turnaround time on updating models and doesn't require a pull request/deploy cycle for silly changes like adding a text option to an array.

To make this change:
1. Add a base Attribute model, with project_name, model_name, attribute_name, and the modifiable attribute fields defined in the template: desc, display_name, match, format_hint. Anything touching the actual model table could not be changed, of course, like indexes, attribute name, type, etc.

2. Update the Magma::Model.json_template function so that it draws from the database for attribute data or uses the Magma::Attribute definition as a default.

graft:
It seems like this issue could be implemented untouched on the road to #104. If we wanted to make it smaller we might slowly phase in the availability of columns in the attribute table (e.g., start with just 'description').

Some notes:

- 1 & 2 described above don't offer any way to update the attribute table. This probably requires the creation of an /update_model API as described in #104, which we might draw out later, but following THIS issue we would only be able to insert/update attribute definitions in the production console.
- Where should the attribute table go? There are two possibilities:
  1. Each project has its own attribute table (that is, `ipi`.`attributes`, the table 'attributes' in the schema 'ipi' in postgres). This allows us to keep the attribute table with each project, but requires separate migrations for each project if we want to update the attribute model.
  2. More likely: there is a single attribute table in some top-level partition shared by all projects. Magma maintains migrations to this table; `bin/migrate` runs these global migrations in addition to the project-specific migrations. There's some code supporting this in the fossilized branch [graft-add-load](/mountetna/magma/tree/graft-add-load) which was meant to fill out #84. In the future (#104) we might expect the project-specific migration code to fall away and only this global migration code remains.

graft:
[...]

PR #123, opened 2020-04-21
Make chosen attribute options editable

alimi:
Per #99, this PR adds a global `attributes` table that stores attribute options `desc`, `display_name`, `match`, and `format_hint`. Options will be fetched from the `attributes` table for a project/model/attribute if it exists; otherwise, we'll continue to use the definition in the model file.

This PR also adds a `global_migrate` command to run database wide migrations. It could be merged into `migrate`, but handling the version flag could be weird. How would we tell if the requested version is for a global migration or a project specific migration?

This is our first pass through this problem -- we're happy for all feedback! 😄

/cc @notmarkmiranda

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,8 @@ Here is an example Magma model:
 
     module Pancan
       class Sample < Magma::Model
-        identifier :name, desc: 'PANCAN id'
-        attribute :mass, type: Float, desc: 'Mass in grams'
+        identifier :name, description: 'PANCAN id'
+        attribute :mass, type: Float, description: 'Mass in grams'
       end
     end
 
--- a/db/migrations/001_add_attributes.rb
+++ b/db/migrations/001_add_attributes.rb
@@ -6,7 +6,6 @@ Sequel.migration do
       String :attribute_name
       String :desc
       String :display_name
-      String :match
       String :format_hint
       DateTime :created_at
       DateTime :updated_at
--- a/db/migrations/001_add_attributes.rb
+++ b/db/migrations/001_add_attributes.rb
@@ -1,0 +1,15 @@
+Sequel.migration do
+  change do
+    create_table(:attributes) do
+      String :project_name
+      String :model_name
+      String :attribute_name
+      String :desc
+      String :display_name
+      String :match
+      String :format_hint
+      DateTime :created_at
+      DateTime :updated_at
+    end
+  end
+end
--- a/db/migrations/001_add_attributes.rb
+++ b/db/migrations/001_add_attributes.rb
@@ -4,7 +4,7 @@ Sequel.migration do
       String :project_name
       String :model_name
       String :attribute_name
-      String :desc
+      String :description
       String :display_name
       String :format_hint
       DateTime :created_at
--- a/db/migrations/001_add_attributes.rb
+++ b/db/migrations/001_add_attributes.rb
@@ -10,6 +10,8 @@ Sequel.migration do
       String :format_hint
       DateTime :created_at
       DateTime :updated_at
+
+      index [:project_name, :model_name, :attribute_name], unique: true
     end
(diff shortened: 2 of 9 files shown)


graft:
>How would we tell if the requested version is for a global migration or a project specific migration?

I think this is already a problem for the existing migrate command, which takes a single version argument and applies it to all projects. It's only been ok because there is a single project in production.

In the branch `graft-add-load` I wrote a version of this command that handles global migrations, with an interface like: `[<project_name> [<version_number>]]`. No arguments would migrate everything to latest; global migrations were run using the project_name 'magma'.

However, since we're hoping to axe the project-specific migrations and only retain the global migrations, i think it's best to keep the two commands separate.

graft:
Magma::Attribute.update_option inserts into the attribute table - what happens after subsequent calls to update_option?

Also while your spec tests update_option for its ability to change the json_template, you also set the `@#{opt}` in fetch_value which I think does not get exercised here.

alimi:
> Magma::Attribute.update_option inserts into the attribute table - what happens after subsequent calls to update_option?

Good catch! In https://github.com/mountetna/magma/pull/123/commits/accaa0d99fcb22248eb6c0cf4344d659938c8f47, I added a unique index on the `attributes` table so it'll only have one row per project/model/attribute. `Magma::Attribute#update_option` will now do an upsert - it'll insert if there's no `attributes` row per project/model/attribute; otherwise, it'll do an update.

> Also while your spec tests update_option for its ability to change the json_template, you also set the @#{opt} in fetch_value which I think does not get exercised here.

I think you're referring to `instance_variable_set("@#{opt}", new_value)` in [`Magma::Attribute#update_option`](https://github.com/mountetna/magma/blob/accaa0d99fcb22248eb6c0cf4344d659938c8f47/lib/magma/attribute.rb#L109). Without setting `@#{opt}`, the test fails. Before the test runs, `spec/spec_helper.rb` starts the `Magma::Server` which will eventually cause all the models (and their attributes) to be loaded into memory.  Attribute options [only get set once when models/attributes are loaded](https://github.com/mountetna/magma/blob/ad8307cc85ce12f4e21c306271d795c13065acdd/lib/magma/attribute.rb#L20). So when the test inserts new values into the `attributes` table, they never show up in the already loaded models. That's why we added `Magma::Attribute#update_option` - it can update the already loaded attribute options be setting `@#{opt}`.

It might be nice to have a test that verifies we're fetching attributes out of the db, but we couldn't find a way around the models already being loaded.

alimi:
Since the match option is more configurable than the other editable attribute options (`desc`, `display_name`, and `format_hint`), we've removed support for editing it in https://github.com/mountetna/magma/pull/123/commits/751b23ac44144589d685e0ec84668fb3d14ca020.

graft:
I have only one tiny request, which is: can we change `desc` to `description`? I've long despised that. Otherwise :ship:

alimi:
@graft yeah, let's change it! I guess we'll also have to update the IPI models to use `description` too?

alimi:
Ok, I replaced `desc` with `description`. `desc` can still be set on model attributes as a fallback. This will let us support stuff like IPI models until they've been updated. (https://github.com/mountetna/magma/pull/123/commits/ed5cbe2a58a7b6d43d7b88492f4f37942008fc7a)

I also left `desc` as the key in `Magma::Attribute#json_template` so the API/client will keep working.

alimi:
I did a little clean up in https://github.com/mountetna/magma/pull/123/commits/3f48b38c7da96e41b3462fa8e4ab21001a3c5863. Before, attribute options would only be set from the database if the attribute was defined with some options in Ruby. Now, attribute options will be set from the database whether or not they're defined in Ruby.

graft:
:+1: :shipit:

Issue #129, opened 2020-04-29
Load attributes from the attributes table


graft:
When Magma boots, it calls Magma#load_models to attach to the database and make the data available via Sequel. The main thing this does is connect to the database, instantiate projects, load the project models, and validate the models.

All of this should continue; we wish to change only Magma::Project#load_project, which calls `require_files('models')`, which includes the models/ directory code in project-specific repos. Instead we wish to achieve the exact same result as this require operation - that is, we have a bunch of model classes < Magma::Model with the correct set of attribute definitions.

To make this change, we might replace `require_files('models')` with `load_project_models`, which reads the model definitions from the attributes table and constitutes a new set of magma models.

Issue #116, opened 2020-03-29
Support table attribute in /update


graft:
The 'table' attribute is the only attribute type that cannot be updated through Magma's API. This is a major weakness that has resulted in a proliferation of back-end data loaders, whose sole responsibility is inserting table data (the various clinical data loaders and the flow_jo loader).

The main reason this situation has persisted is the need to delete data. While it is probably the case that individual table elements could be successfully orphaned through the /update API (by setting their parent to null), this is probably a bug and no current code attempts to exploit this deficiency. Barring this there is no facility to actually remove records from Magma; therefore, in order to replace the contents of a table attribute we need to delete the old contents via the backend, where we can use Sequel to do it.

This remains the main special ability of all of the data loaders: the ability to remove the old data they are meant to replace. This is insidious! We shouldn't let this tiny requirement chain us to an entire system of back-end data loading.

The solution is simple: allow the /update interface to send tables. When someone posts a new table to a record (or records), the old contents of that record's table attributes (belonging to the associated model) are first removed (or following #115 simply orphaned) before the new data is inserted. There is some dispute to be had about the syntax of the insertion, which I will reserve for later.

This change would allow us to discard all of the back-end data loaders and instead allow the same loaders to run as clients to Magma.

graft:
Closed by mountetna/monoetna#90 and #197

PR #132, opened 2020-04-30
Define Magma::Model.load_attributes

alimi:
This builds on https://github.com/mountetna/magma/pull/123. `Magma::Model.load_attributes` pulls a model's attributes from the database and defines them on the model using the newly added `type` column.

With `Magma::Model.load_attributes`, we'll no longer need to load attribute options when initializing `Magma::Attribute` objects. This will also give us a little performance boost since we'll get all a model's attributes in one database query. However, we'll no longer be able to merge attributes defined both in Ruby and the database.

This PR also makes more `Magma::Attribute` options editable, but `match` is still not supported. It also makes `loader` and `link_model` editable which might not be desired. I can remove it--just lemme know 😄.

I'm also not sure where we should call `Magma::Model.load_attributes`. I could see make adding it explicitly in models like

```ruby
module Labors
  class Monster < Magma::Model
    parent :labor

    identifier :name, type: String
    # …

    load_attributes
  end
end
```

so the database attributes take precedence over anything defined in the Ruby files. Alternatively, I could see calling it after the definition to guarantee it's always called.

```ruby
class Magma
  # …
  class Model
    class << self
      # …
      def load_attributes
        # …
      end

      load_attributes
    end
  end
end
```

But this would let Ruby attribute definitions override the database definitions. It might not be a big deal since we want to move away from Ruby definitions anyways. I'm curious on what y'all think.

/cc @notmarkmiranda @graft


graft:
You could have Magma call it in load_models or whatnot, after the models themselves have loaded, this would be in line with #129.

alimi:
Cool, that makes sense. I added it to the end of `Magma::Project#load_project` so the db attributes get set after the Ruby attributes. https://github.com/mountetna/magma/pull/132/commits/bc5c12734a573d65a3d95661430b363ee68e1ef3

lib/magma/project.rb
```diff
@@ -47,6 +47,20 @@ def load_project
...
+
+    def load_model_attributes
+      model_attributes = Magma.instance.db[:attributes].
+        where(project_name: @project_name.to_s).
+        to_a.
+        group_by { |attribute| attribute[:model_name].to_sym }
+
+      model_attributes.each do |model_name, attributes|
+        model = models[model_name]
+        model.load_attributes(attributes)
```
graft, line 62: might this be null?

lib/magma/project.rb
```diff
@@ -47,6 +47,20 @@ def load_project
         require_files('loaders')
         require_files('metrics')
       end
+
+      load_model_attributes
+    end
+
+    def load_model_attributes
+      model_attributes = Magma.instance.db[:attributes].
```
graft, line 55: Is the intent here to save database calls? Seems like it would be clearer to put this back in `Magma::Model`

lib/magma/project.rb
alimi: Are you saying the model might be null? It's possible, but seems unlikely. I can add a check though to make sure this doesn't blow up if there isn't a model for the attribute.

lib/magma/project.rb
alimi: Yeah, that's why I moved it. For a project like IPI that has 17 models, doing the query upfront saves us 16 trips to the database. As a we add more projects, those queries would add up quickly. I was torn on moving it, but the trade off seemed worthwhile.

db/migrations/002_add_columns_to_attributes.rb
```diff
@@ -0,0 +1,17 @@
+Sequel.migration do
+  change do
+    alter_table(:attributes) do
+      add_column :type, String, null: false
+      add_column :hidden, "boolean", default: false
+      add_column :read_only, "boolean", default: false
+      add_column :unique, "boolean", default: false
+      add_column :index, "boolean", default: false
+      add_column :loader, String
+      add_column :restricted, "boolean", default: false
```
alimi, line 10: I updated the columns per https://github.com/mountetna/magma/issues/99#issuecomment-621507560, but this includes a few more that aren't listed there. Should we keep `unique`, `index`, and `loader`?

db/migrations/002_add_columns_to_attributes.rb
```diff
@@ -0,0 +1,17 @@
...
+
+      set_column_not_null :project_name
+      set_column_not_null :model_name
+      set_column_not_null :attribute_name
```
alimi, line 14: I made `project_name`, `model_name`, `attribute_name`, `type` not null because they have to be set for this to work long term.

lib/magma/attribute.rb
```diff
@@ -1,14 +1,25 @@
...
+      :unique
+    ]
+
+    attr_reader :name, :loader, :validation, :format_hint, :unique, :index, :restricted, :description
 
     class << self
       def options
-        [:description, :display_name, :hide, :readonly, :unique, :index, :validation,
-:format_hint, :loader, :link_model, :restricted, :desc ]
+        [:description, :display_name, :hidden, :read_only, :unique, :index, :validation,
```
alimi, line 21: `hide` and `readonly` were renamed to `hidden` and `read_only` per https://github.com/mountetna/magma/issues/99#issuecomment-621507560.

lib/magma/attribute.rb
```diff
@@ -26,6 +37,10 @@ def database_type
       nil
     end
 
+    def type
+      @type ||= self.class.name.match("Magma::(.*)Attribute")[1].underscore
```
alimi, line 41: I added this to make `#update_option` a little easier to work with.

https://github.com/mountetna/magma/blob/e983166a3100529a81ddb1e3191509e54f7fc287/lib/magma/attribute.rb#L115-L134

It works but it feels a little brittle since it's derived from [`Magma::Model::ATTRIBUTE_TYPES`](https://github.com/mountetna/magma/blob/453cf160ffc7207a4ac7e25d2bb027ebc683e179/lib/magma/model.rb#L9-L12) and there's nothing really tying them together. It might also be a little confusing with `database_type`.

lib/magma/attribute.rb
```diff
@@ -86,10 +101,6 @@ def display_name
       @display_name ||= name.to_s.split(/_/).map(&:capitalize).join(' ')
     end
 
-    def description
-      @description || @desc
-    end
```
alimi, line 91: `@desc` was hanging around as a fallback but we don't need anymore because IPI is already using `description`.

lib/magma/project.rb
alimi: @graft can you clarify what you mean here?

graft: If the attributes table (+ a models table if we add one) ever diverge from the database or the class definition we would have a conflict here (if, e.g., we remove one of the models from the magma-ipi repo, but there is still a definition for it in the attributes table). When we boot and try to attach those attributes to the non-existent model, this will fail; we should guard against this possibility of divergence by raising the appropriate error here to warn about the mismatch.

db/migrations/002_add_columns_to_attributes.rb
graft: I think `loader` will shortly be deprecated anyway; there is really no way to run them using the existing methodology (as hooks in the upload process) after the magma/metis changes. We need #116 to develop an alternative, but I think we should just abandon it for the moment. `unique` and `index` are okay, although I'm wondering about column proliferation now...

lib/magma/attribute.rb
graft: I like `attribute_type` here. At some point I think we should build `ATTRIBUTE_TYPES` instead of hard-coding it; we could use this method to do so, which should tie them together nicely.

graft: When `type` moves into `EDITABLE_OPTIONS` what happens to line 127 here?

alimi: > I like `attribute_type` here. At some point I think we should build `ATTRIBUTE_TYPES` instead of hard-coding it; we could use this method to do so, which should tie them together nicely.

Cool, I'll rename it to `attribute_type`. I was also playing around with dynamically building `ATTRIBUTE_TYPES`, but I backed off because I wasn't sure we wanted to go in that direction. I'll give it another try.

> When `type` moves into `EDITABLE_OPTIONS` what happens to line 127 here?

Hmm...I'm not following. When does `type` move into `EDITABLE_OPTIONS`?

db/migrations/002_add_columns_to_attributes.rb
alimi: Ok, I made `loader` non-editable in 5b00d86fd623b6cebaa301829c4980a7a4290539.

lib/magma/attribute.rb
alimi: I made `attribute_type` a class method in cd71a7695caaa35b2d198d449296495258f994b0 so we can dynamically generate the `Magma::Model` attribute methods.

https://github.com/mountetna/magma/blob/cd71a7695caaa35b2d198d449296495258f994b0/lib/magma/model.rb#L10-L17

lib/magma/project.rb
alimi: Gotcha, thanks for explaining 😄. If this happens, we'll raise a `Magma::Project::LoadError`. See 94eed4c2e6939bbffa71b251ee1687eda32002cd.

alimi:
@notmarkmiranda heads up I added a test here for loading link models from the attributes table.

https://github.com/mountetna/magma/blob/cfc38653df3286c57121460c527f37617c04954d/spec/magma_project_spec.rb#L40-L56

I had this test before https://github.com/mountetna/magma/pull/134, but it was just using `link_model` instead of `link_model_name`. It wasn't too hard to resurrect 🧟‍♂️

alimi:
@graft I have a couple of questions

1. With this change, if a model has an attribute defined both in Ruby and the attributes table, the Ruby definition will be completely replaced with the db definition. Is this OK? I'm not sure if we'll want to be able to have attributes in both places while we migrate to having everything in the db.
2. Can you clarify what you mean in this comment: https://github.com/mountetna/magma/pull/132#discussion_r426116616?

lib/magma/project.rb
```diff
@@ -47,6 +49,31 @@ def load_project
...
+        where(project_name: @project_name.to_s).
+        to_a.
+        group_by { |attribute| attribute[:model_name].to_sym }
+
+      model_attributes.each do |model_name, attributes|
+        model = models[model_name]
+        raise load_error(model_name, attributes) unless model
```
graft, line 64: I would prefer to move the logic in `#load_error` into `LoadError` so that we can `raise LoadError` here, not being able to see what error class you are raising is too opaque.

graft:
1. It seems like this is okay, since we will have the choice of when we replace the Ruby attributes with the db attribute. We might want to write some attribute-auditing code to identify discrepancies before we unhook the models fully.

2. I don't know what I meant any more. I think I may have been confused about the type being included on updates, but it is only included in the intial insert.

lib/magma/project.rb
alimi: Ok, I moved it in https://github.com/mountetna/magma/pull/132/commits/b94e8b1f5f927f7f9e8acb2b26d0275c47821a04. I also renamed it to `AttributeLoadError` because it better represents what the error is.

Issue #125, opened 2020-04-29
Clarify types


graft:
Now that we have safely hidden away the database type of Magma, what *is* the type system Magma exposes?

Currently this type system is defined in the interface of Magma::Model, which gives a number of helper methods that allow the construction of attributes. To enumerate, they are:
- value types, broken into:
  - basic types - `string integer boolean date_time float`
  - file types - links to files on metis/s3 etc. `file image`
  - json types - data formats stored in a json column `match matrix`
  - `identifier` - the unique string identifier for the record

- link types, broken into:
  - one-to-many -  `collection table`
  - one-to-one - `child parent link`

Per 'load attributes from attributes table` there should be a single "type" column with an enumerated value that clearly determines the type of an attribute, to replace the interface in `Magma::Model`. This means essentially that all of the attributes should be defined as they are here: https://github.com/mountetna/magma/blob/d6764d31e6e64b9447f6480e6af58cf14c3db0e1/lib/magma/model.rb#L10, that is, there is a single class corresponding to the attribute type which can be instantiated to setup the attribute correctly.

However, a number of unruly attributes continue to resist this pattern and do not have their own attribute classes, namely `identifier`, `parent`, and `link`. Probably these can simply be given their own classes inheriting from the parent class each currently relies on (`StringAttribute` and `ForeignKeyAttribute`).

graft:
Closed, methinks, by various changes along the migration API roadmap.

PR #133, opened 2020-05-01
Refactor identifier, parent, and link attributes

notmarkmiranda:
#125 

This will refactor identifier, parent, and link to match the other attribute patterns. The only holdover was that the previous #parent method was a getter and setter method. There is an instance variable set in the define method group in order to set the parent on the model.

### DIFF
--- a/lib/magma/model.rb
+++ b/lib/magma/model.rb
@@ -6,12 +6,16 @@ class Magma
   Model = Class.new(Sequel::Model)
    
   class Model
+    ATTRIBUTES = [
+      :string, :integer, :boolean, :date_time, :float, :file, :image, 
+      :collection, :table, :match, :matrix, :child, :identifier, :parent, :link
+    ].freeze
     class << self
-      %i(string integer boolean date_time float file image collection table match matrix child foreign_key identifier parent link).each do |method_name|
+      ATTRIBUTES.each do |method_name|
         define_method method_name do |attribute_name=nil, opts={}|
           klass = "Magma::#{method_name.to_s.capitalize}_attribute".camelcase.constantize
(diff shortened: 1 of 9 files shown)


graft:
Very nice. A few remarks:

1) Could we use `parent_model_name` instead of `parent` or `parent_model`? The latter suggests a `Magma::Model` object to me, the former is ambiguous.

2) In line 10 in `model.rb` we define the array of types. It seems a bit ignominious to have this canonical type definition nestled in this iterator. Options might be: 1) move it to a constant or 2) collect it from the attribute class names. But per (3) below, not all attribute classes might be exposed, so we would probably need to define some sort of flag indicating that an attribute class also represents an exposed type. The advantage of (2) is not having to maintain a list in model.rb, though (1) is obviously easier to implement and maybe clearer in the end.

3) Also in line 10 in `model.rb`, while I might be mistaken, I don't think `foreign_key` should actually be an exposed type, as it is just a base class for `parent` and `link` types.

lib/magma/model.rb
```diff
@@ -6,10 +6,15 @@ class Magma
...
    
   class Model
+    ATTRIBUTES = [
+      :string, :integer, :boolean, :date_time, :float, :file, :image, 
+      :collection, :table, :match, :matrix, :child, :identifier, :parent, :link
+    ].freeze
     class << self
-      %i(string integer boolean date_time float file image collection table match matrix child foreign_key).each do |method_name|
-        define_method method_name do |attribute_name, opts={}|
+      ATTRIBUTES.each do |method_name|
```
graft, line 14: I might rename this as `ATTRIBUTE_TYPES` and `attribute_type` for clarity.

graft:
:+1:

Issue #127, opened 2020-04-29
Support link_model in attributes table


graft:
The LinkAttribute takes an option `link_model` which specifies what the model it links to is - this allows us to separate the model name from the link. We need to be able to specify this option in the attribute table.

PR #134, opened 2020-05-06
Add support for link_model on attributes table

notmarkmiranda:
#127 

This will add a `link_model` attribute to the `attributes` table. It also adds `:link_model` to the EDITABLE_OPTIONS of attribute.


graft:
Hmm, is this all it takes now to have the link_model load from the attribute table? If so, very cool. Would be nice to have some specs validating this, not sure how to set that up...

alimi:
We could cover it in one or all of these tests:

https://github.com/mountetna/magma/blob/b9ce3a1028847ece9239d841e5fe0236a0ee8024/spec/magma_attribute_spec.rb#L5-L43

Everything in `EDITABLE_OPTIONS` should behave the same though so I'm not sure we want to test every option.

graft:
I think you are overwriting the behavior of `Magma::Link#link_model`, a module which is included into some Magma attributes. I have some regrets about writing this module, but this method is used extensively throughout Magma so it would be pretty tough to change. The method uses the instance variable `@link_model`which is set on the model, but itself returns the model class, not a string. This behavior is different from the `attr_reader :link_model` you defined, which returns a string. This doesn't explode tests because `include Magma::Link` appears to overwrite the `attr_reader :link_model`, but it seems like this problem should be avoided by renaming `@link_model` to `@link_model_name` and changing the attr_reader (and maybe the column name in the table?) appropriately.

Since this is the main thing we to ensure we can do (successfully link to another model), it might be nice to split this out from the other options test, and to fully exercise the behavior of `Magma::Link#link_model` starting from the attribute fixture, and using some sort of link attribute (like child or collection attribute) instead of a generic `Magma::Attribute`. Also, I think `link_model_name` should be formatted like `model_name` (that is, snake_case) rather than holding a ruby class name.

Also: I think somewhere on the way this thing should be raising an exception if the link_model_name does not exist in the project - maybe using `Magma.instance.get_model`? A test proving this would also be useful.

.ruby-version
```diff
@@ -0,0 +1 @@
+2.5.7
```
alimi, line 1: Does this file need to be committed?

db/migrations/002_add_link_model_to_attribute.rb
```diff
@@ -0,0 +1,9 @@
+Sequel.migration do
```
alimi, line 1: Now that https://github.com/mountetna/magma/pull/140 has been merged, there's already a migration version 002 ([`db/migrations/002_add_validation_to_attributes.rb`](https://github.com/mountetna/magma/blob/4880aaf8c2720465a84f8fed52a8dc5179f17316/db/migrations/002_add_validation_to_attributes.rb)). Sequel won't let us have two migrations with the same version, so this file will have to be renamed to 003.

lib/magma.rb
```diff
@@ -75,6 +75,7 @@ def validate_models
 
         # Check reciprocal links
         model.attributes.each do |att_name, attribute|
+          #require 'pry'; binding.pry if att_name == :labor && attribute.nil?
```
alimi, line 78: This line can probably be removed.

notmarkmiranda:
Merging this to avoid scope creep and leaving the other refactors for another time. Not wanting to block progress on other things.

Issue #126, opened 2020-04-29
Add/repair validations for attributes


graft:
The validation system runs a number of different kinds of validation, one of which is attribute validation (`Magma::Validation::Attribute`, not to be confused with `Magma::Attribute::Validation`, although you will). Essentially, this will ensure that each attribute thinks the value being passed in is valid according to the requirements of the attribute (whatever they are). The corresponding attribute validation is performed by each attribute's validation class (e.g. `Magma::IntegerAttribute::Validation`), if these exist. Apparently, because of the way constants work, if they do not exist it will default (probably incorrectly) to `Magma::Attribute::Validation`.

This class is the sole place where the `match` parameter of the attribute is used (other than an obscure usage by the `matrix` attribute); it seems as if the usage of this parameter as written in `Magma::Attribute::Validation` is really only appropriate to string attributes (i.e., match against a regexp and match against an array of strings) and could (should) be renamed to `Magma::StringAttribute::Validation`.

The purpose of the match parameter on the attribute is to provide a way to lightly configure these attribute-level validations. A few of the attribute classes have their own custom validations, but there are some holes (notably, float/integer/date_time should have range validations on them).
In the future we'd also like to move these to a json column in the attributes table.

The work:
0) Rename `Magma::Attribute#match` to `validation`
1) Create validation classes for `IntegerAttribute`, `FloatAttribute`, `DateTimeAttribute` which use the `@validation` parameter to validate ranges for these values.
2) Move `Magma::Attribute::Validation` to `Magma::StringAttribute::Validation`
3) Investigate if other validations make sense for other types.

Lastly, once this is done:
4) Add a `validation` json column to populate the attribute.  Note that previously we used the ruby class to determine how the matcher works (i.e., Array vs. Regexp); this might be a bit confused in JSON, which has no Regexp type. While we might equivocate and try to make a JSON string pass as a regexp or some other nonsense, probably we should be more explicit. Conveniently, the `MatchAttribute` actually already defines a JSON format for holding a match configuration that works against a number of different types; we should probably re-use this here.

alimi:
> 1. Create validation classes for IntegerAttribute, FloatAttribute, DateTimeAttribute which use the @match parameter to validate ranges for these values.

@graft should "use the `@match` parameter" be "use the `@validation` parameter" since `match` will be renamed to `validation` in step 0?

alimi:
Will the new validation parameter still need to support procs like match currently does?

[...]

Issue #128, opened 2020-04-29
Clean up Magma::Attribute#json_template


graft:
Ye olde `#json_template` method on `Magma::Attribute` is an ancient piece of Magma code, and probably getting crusty now that we are attempting to clean up the structure of the attributes themselves. The main reformation we wish to make is to bring it in line with the model specified in #99. The changes:
1) Consolidate 'match' and 'options' following #126 - this will mostly make Timur upset since it uses some of these values to control its AttributeViewer component.
2) Remove the odd 'model_name', which appears to only be necessary for link attribute types. However, this should be reported somehow once we resolve how to store this value in the attribute table, see #127.

alimi:
> 1) Consolidate 'match' and 'options' following #126 - this will mostly make Timur upset since it uses some of these values to control its AttributeViewer component.

How are you thinking to consolidate `match` and `options`? Would we combine them into a single `validation` item?

graft:
I think I wrote this before we discussed better strategy for not breaking Timur - a non-breaking way to do this would be to maintain match and options for now but add in the `validation` item, which would just include the validation JSON. We can patch Timur to use this json, then retire the 'match' and 'options'.

PR #135, opened 2020-05-06
Update attribute validations

alimi:
This PR addresses the first few things outlined in #126.

- [x] Rename `Magma::Attribute#match` to `validation`
- [x] Create validation classes for `IntegerAttribute`, `FloatAttribute`, `DateTimeAttribute` which use the `@validation` parameter to validate ranges for these values.
- [x] Move `Magma::Attribute::Validation` to `Magma::StringAttribute::Validation`
- [ ] Investigate if other validations make sense for other types.
- [ ] Add a validation json column to populate the attribute.

It doesn't add validation to the attributes table, but it takes a step in that direction by making the Ruby validation definitions hashes/JSON. For example, regex validation on a string attribute looks like

https://github.com/mountetna/magma/blob/422807062c90a2bea248a3d48b8ef26468832c8b/spec/labors/models/monster.rb#L6

I did this to lay the foundation for moving it to the attributes table. I also wanted to get feedback on the way things look now before we take that next step.

I also haven't looked to see if we should add more attribute validations.

### DIFF
--- a/lib/magma.rb
+++ b/lib/magma.rb
@@ -2,6 +2,7 @@ require 'sequel'
 
 require_relative 'magma/project'
 require_relative 'magma/validation'
+require_relative 'magma/validation_object'
 require_relative 'magma/loader'
 require_relative 'magma/migration'
 require_relative 'magma/attribute'
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -27,8 +27,7 @@ class Magma
     end
 
     def validation_object
-      return unless @validation
-      @validation_object ||= validation_class.new(*validation_arguments)
+      @validation_object ||= Magma::ValidationObject.build(@validation)
     end
 
     def json_template
@@ -39,8 +38,8 @@ class Magma
         attribute_class: attribute_class_name,
         desc: description,
         display_name: display_name,
-        options: validation_object.is_a?(Array) ? validation_object : nil,
-        match: validation_object.is_a?(Regexp) ? validation_object.source : nil,
+        options: validation_object.options,
+        match: validation_object.match,
         restricted: @restricted,
         format_hint: @format_hint,
         read_only: read_only?,
@@ -167,13 +166,6 @@ class Magma
       else
         self.class.name
       end
-
-    def validation_class
-      @validation[:type].constantize
-    end
-
-    def validation_arguments
-      [@validation[:value]]
     end
 
     class Entry < Magma::BaseAttributeEntry
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -3,11 +3,11 @@ class Magma
     DISPLAY_ONLY = [:child, :collection]
     EDITABLE_OPTIONS = [:description, :display_name, :format_hint]
 
-    attr_reader :name, :loader, :match, :format_hint, :unique, :index, :restricted
+    attr_reader :name, :loader, :validation, :format_hint, :unique, :index, :restricted
 
     class << self
       def options
-        [:description, :display_name, :hide, :readonly, :unique, :index, :match,
+        [:description, :display_name, :hide, :readonly, :unique, :index, :validation,
 :format_hint, :loader, :link_model, :restricted, :desc ]
       end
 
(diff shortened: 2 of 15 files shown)


graft:
This is some nice work. Two remarks:

1) Since there is so much type checking of `validation_object` in the Validation class, you could pass the entire @validation parameter into the validation class instead of using `validation_arguments` and `validation_object` in `Magma::Attribute`.

2) There is a nearly-identical usage of this validation format here: https://github.com/mountetna/magma/blob/b9ce3a1028847ece9239d841e5fe0236a0ee8024/lib/magma/dictionary.rb#L46 - it might be nice to use a single class for both of these formats.

alimi:
> 1) Since there is so much type checking of `validation_object` in the Validation class, you could pass the entire @validation parameter into the validation class instead of using `validation_arguments` and `validation_object` in `Magma::Attribute`.

I like this idea, but there are a few places outside the validation classes that need to know what the validation object is:

https://github.com/mountetna/magma/blob/422807062c90a2bea248a3d48b8ef26468832c8b/lib/magma/attribute.rb#L42-L43

https://github.com/mountetna/magma/blob/422807062c90a2bea248a3d48b8ef26468832c8b/lib/magma/attributes/matrix.rb#L74-L82

https://github.com/mountetna/magma/blob/422807062c90a2bea248a3d48b8ef26468832c8b/lib/magma/query/predicate/matrix.rb#L23-L25

https://github.com/mountetna/magma/blob/422807062c90a2bea248a3d48b8ef26468832c8b/lib/magma/query/predicate/matrix.rb#L36

I'm not sure it makes sense to push that logic into the validation classes when we need it elsewhere.

graft:
I think you are right about not hiding the object inside the validation class. How about a `ValidationMatcher` class to wrap that `{ type, value }` format? So you could do:

     options: validation_object.options, 
     match: validation_object.match, 
and

    validation_object.matches?(value)
and probably reuse this in `MatchAttribute`.

alimi:
Actually, it looks like the `else` clause catches strings?

graft:
You seem to be correct about `{ type: 'String' }`. However, this makes me cognizant of how I am probably leading you astray sending you after the MatchAttribute, because I don't really know the answer! I implemented this and it is available, but we never wrote a dictionary for the ipi project, so it's difficult to know what the actual usage would look like.  Since it's not being used in production it seems low-value to work on it; also, we'll be forced to re-encounter this if we actually want to add dictionary validation back to the models via the API, at which point we might unify MatchAttribute with the attribute validation interface.

alimi:
Yeah, the else clause catches strings and numerics. I came across them in `spec/validation_spec.rb`.

Is there anything else to work on for this PR or can we ship it? I'm thinking to do add the json column to the attributes in another PR so #128 doesn't get blocked.

graft:
Looks like it needs a rebase; i still think a matcher class to wrap the validation format has value for this PR, but we should restrict scope as much as possible, I'm getting frightened by all the feature dev creeping into the migration roadmap.

alimi:
Ok, I added `Magma::ValidationObject` in https://github.com/mountetna/magma/pull/135/commits/5dac6ac2a50adb983d3f958328a6722057af4baf. There's a different `Magma::ValidationObject` class for each of the validation types (`Magma::ArrayValidationObject`, `Magma::RangeValidationObject`, etc.).

https://github.com/mountetna/magma/blob/5dac6ac2a50adb983d3f958328a6722057af4baf/lib/magma/validation_object.rb#L1-L92

It's nice to have all that logic in one place. It also helps us eliminate a lot of the type checking that was going on. I added `Magma::NullValidationObject` to get us away from some nil checks too.

I think it makes sense to leave `MatchAttribute` as is for now.

alimi:
I also brought back proc support for regex validations in https://github.com/mountetna/magma/pull/135/commits/9a241a0019988f374845e508ecbeaab6c20aded3 so the IPI project can continue to build regexes with database values.

graft:
Looks nice! Is it possible to remove the empty `class Validation < BaseAttributeValidation` now (since I think it should default to BaseAttributeValidation)? Otherwise seems fine to :shipit: 

The proc support is nice since otherwise we would be breaking the validation in production and would have to amend it. Going forward (adding the json column to the db) I think we'd have to abandon this, though?

alimi:
Ah, good call! I removed the empty validation classes in https://github.com/mountetna/magma/pull/135/commits/1e33bbe60bed584a8e03acb9e0248030f7420348.

Yeah, I don't think we'll be able to support procs in the attributes table but we can deal with it when we get there.

alimi:
I pushed up a test fix in https://github.com/mountetna/magma/pull/135/commits/5425659434673d108956f1ed1abee71fc6594226. I accidentally made the validation spec flaky, but it's fixed now.

PR #136, opened 2020-05-08
Use old attr class names in json_template

alimi:
Since we've added new attribute classes, such as `Magma::BooleanAttribute` and `Magma::LinkAttribute`, `Magma::Attribute#json_template` has new values for `attribute_class`. This breaks the API for consumers like timur.

To keep the API consistent, we'll continue using the old attribute class names in `#json_template`. 

This should get timur back to a working state without needing any changes. Since I'm newer to timur, I could use another set of eyes to verify things work 😄


graft:
:+1: In the future we can add `attribute_type` to expose the magma type (integer/boolean/matrix, etc), make timur work with that, then phase this out.

PR #137, opened 2020-05-11
Issue51 files direct to metis

coleshaw:
@graft , here is what I have so far -- do you think I'm going in the right direction? From a high-level perspective, in addition to updating your `40-use-loader-for-update` branch against `master`, I've added a method to `MetisStorage` class to generate an HMAC-signed "copy" URL and also inserted a new method in the workflow for `update#action` to do a POST against the copy URL on Metis.

I also have a couple of additional, specific questions:

* [update.rb](https://github.com/mountetna/magma/compare/master...coleshaw:issue51-files-direct-to-metis?expand=1#diff-efe99846a754ff3caa1a4c083a444eb3R77), when I'm instantiating a new `Etna::Client`, do I pull the `token` field from the config, as shown? Or should that get populated some other way?
* [update.rb](https://github.com/mountetna/magma/compare/master...coleshaw:issue51-files-direct-to-metis?expand=1#diff-efe99846a754ff3caa1a4c083a444eb3R82), I feel like this code to check if a revision includes a file attribute and getting the file path are brittle. Thoughts on a better way to accomplish this?
* [storage.rb](https://github.com/mountetna/magma/compare/master...coleshaw:issue51-files-direct-to-metis?expand=1#diff-ee75dc68c65abf60f094ca03a7e95a0f), the new `copy_url` method assumes the bucket name on Metis is `magma` -- is that a safe assumption? Should that be a different value or handled in a different manner?

Also random question: When testing via Timur, where would I see the file links in the Metis database? In `files` table?

Thanks!


lib/magma/server/update.rb
```diff
@@ -55,4 +57,36 @@ def load_revisions
...
+  def update_any_file_links
+    # Here, if there are any File attributes in the revisions and
+    #   the storage is configured to use Metis instead of AWS,
+    #   we'll send a request to the Metis "copy" route, using the
+    #   Etna::Client.
+    if Magma.instance.config(:storage).fetch(:provider).downcase == 'metis'
+
+      # Etna::Client.new throws an exception if the host config includes protocol,
+      # i.e. https://metis-dev.etna-development.org will throw an exception
+      raise Etna::BadRequest, 'Storage host should not include protocol' if
```
graft, line 71: This error might be raised when you load storage; the client probably doesn't need to know about this.

lib/magma/server/update.rb
```diff
@@ -55,4 +57,36 @@ def load_revisions
...
+        Magma.instance.config(:storage).fetch(:host).start_with? 'http'
+
+      # Do we get the token from config?
+      client = Etna::Client.new(
+        "https://#{Magma.instance.config(:storage).fetch(:host)}",
+        Magma.instance.config(:token) ? Magma.instance.config(:token) : 'token')
+
```
graft, line 78: The token (which is the user's JWT token) should come from the request, you can see in https://github.com/mountetna/etna/blob/c3240af177b26f50fa228304b7e955b7e09f4c43/lib/etna/auth.rb#L68 how to grab it. When we pass this on to metis, we are making the request to Metis as whatever user made the request to magma.

lib/magma/server/update.rb
```diff
@@ -55,4 +57,36 @@ def load_revisions
...
+      @revisions.each do |model, model_revisions|
+        model_revisions.each do |revision|
+          if revision.is_a?(File)
+            # copy_url = Magma.instance.storage.copy_url(
+            #   @project_name,
+            #   revision. # Where do I just pull out the path from?
+            # )
+
+            # Use the Etna::Client to make a POST call to Metis's copy_url?
```
graft, line 89: I'm not sure how well Etna::Client will play with a full URL, since its helpers expect the route name and try to build the request path for you. So we need to do something like:
  1) Build the route path for the copy request (probably just with a string template)
  2) Use Etna::Hmac to compute the hmac params for that path
  3) Use Etna::Client to request that path + hmac params (probably circumventing the copy route's helper method and using Etna::Client#post or something)

lib/magma/server/update.rb
```diff
@@ -22,6 +22,8 @@ def action
 
     load_revisions if success?
 
+    update_any_file_links if success?
+
```
graft, line 27: I think the metis file stuff must happen before the load request to the database, because if metis replies with a 403 or something, Magma should not perform the db update. On the other hand, the load itself sometimes fails because of poor upstream validation, resulting in inserting the wrong type, duplicate keys, etc. Hard to know how to resolve this conundrum.

graft:
Looks along the way, I tried making some comments. I'm having a bit of trouble following because of all the rebasing going on. It seems kind of problematic that all of this stuff is happening on top of the update/loader changes, I'd really prefer to merge those - but those changes are pretty much impossible to merge without the subsequent Metis changes, so I guess rebasing is where we are at until this is ready to go. (on that note - can we work in a branch on this repo rather than in a fork?)

One general thing I am noticing: You are inserting a lot of case statements about the provider being Metis, because I wrote the storage class to appear as if we can switch seamlessly between AWS and Metis as our file storage. However, this is not really true - in the current storage we are sending files straight to magma, using Metis storage we are sending files straight to Metis. This is fairly different in concept and can't really be shifted as a config flag. We should probably just acknowledge in the code that Metis is Magma's storage provider and get rid of all of this case logic. In the future if we want to enable a direct AWS backend for Magma we can build that back in, but really we are breaking the current storage system with this update/loader stuff.

coleshaw:
Thanks for the feedback, I'll give it another pass.

I agree it's confusing looking at this on top of the loader work ... it would definitely be cleaner if we could see just these changes. What are your thoughts about re-doing the PR but just against the loader branch, and we rebase against master as the final step? I can then also move the work to the `mountetna` org instead of a fork.

Regarding 100% commitment to Metis instead of AWS, that brought up a couple of questions:

1) Is there any way to tie that configuration change into the build and deploy process? It seems like an easy thing to forget, but it's also a breaking change in how Magma works, so a way to tie in the change to `config.yml` seems like it would be good.

2) Are there any migrations required for existing data, to get them into the new "Metis" format vs. old AWS format?

lib/magma/server/update.rb
coleshaw: Hm....interesting. Will noodle on it.

lib/magma/server/update.rb
coleshaw: Okay, sounds good. So I don't need a separate `copy_url()` method that computes the full URL path + HMAC values.

graft:
I think you are right about not rebasing, I was trying to think about keeping up-to-date with that and it seems nightmarish. We can merge to `40-use-loader-for-update`, I think that works, but we should probably do some review on those changes first, because they haven't really had any.

Regarding deploying this stuff, that is a good question. It involves a major movement of data between S3 and Metis, so there's little chance we can do it in one seamless deploy with no downtime. What I was imagining:
1) Copy the data from S3 into Metis.
2) Copy the values of file attributes from Magma - use the names here and the ones from S3 in (1) to determine which file on Metis should be linked to which record/file attribute.
3) Introduce the new changes. Zero out the existing file_attribute column on Magma.
4) Use the map in (2) to update magma with the file locations on Metis.

I think you're also right about the brittleness of using a fixed path for the copy route also, as that might change on metis. I realized however that Etna::Client has a private `#route_path` method that lets you construct the path using up-to-date routes from Metis. Still a little brittle in that the param names it uses might change, but better than just a fixed string template.

Regarding the bucket name 'magma', this is an interesting question we should guarantee somehow, I wrote it up [here](https://github.com/mountetna/metis/issues/66).

As for how to see stuff on Metis when making updates via Timur, yes, you should see a file appear in the appropriate Magma bucket - you can look in the Metis::File model for it in the console.

lib/magma/server/update.rb
coleshaw: Is there another exception that I should raise here, to let a deployment person know that storage is configured incorrectly? Or are we just relying on outside knowledge of the acceptable format?

coleshaw:
So I'm going to close this PR and move to using the `40-use-loader-for-update` branch as the basis for this work. Will also file a separate PR on that branch to review those changes (though I think that will look messy due to the conflicts with the current work on migration API).

PR #139, opened 2020-05-12
Issue51 copy file

coleshaw:
@graft , can you give this another look? One thing I wasn't sure about in the [test stubs](https://github.com/mountetna/magma/compare/40-use-loader-for-update...issue51-take2-use-metis-storage?expand=1#diff-f999d89edde83b171d0e86e801229cdcR12) was the exact format of the returned "routes" that get fed into `Etna::Client`. The definitions in Metis show:

```
/:project_name/file/copy/:bucket_name/*file_path
```

But I don't think the route parser handles `*file_path` correctly, and I'm not exactly sure how it gets populated "in real life" in Etna (i.e. `*file_path` or `:file_path`). Is there a way I can investigate that locally? I've tried going into a Magma console (`bin/magma console`), but can't stand up an Etna::Client because it can't verify SSL certs.

### DIFF
--- a/lib/magma/server/update.rb
+++ b/lib/magma/server/update.rb
@@ -166,5 +166,8 @@ class UpdateController < Magma::Controller
       hmac.url_params[:query]).map { |k,v| [k.to_sym, v] }.to_h
 
     client.send('post', hmac.url_params[:path], copy_params.merge(hmac_params))
+  rescue Etna::Error => e
+    log(e.message)
+    @errors.concat([e.message])
   end
 end
--- a/lib/magma/server/update.rb
+++ b/lib/magma/server/update.rb
@@ -83,16 +83,15 @@ class UpdateController < Magma::Controller
     @revisions.each do |model, model_revisions|
       model_revisions.each do |revision|
 
-        revision.attribute_names.each do |attribute|
-          if is_file_attribute(revision.model, attribute)
-
-            if !revision.to_loader[attribute]  # nil
-              remove_copy_on_metis(revision, attribute)
-            elsif revision.to_loader[attribute][:location].start_with? 'metis:'
-              copy_file_on_metis(revision, attribute)
-            elsif revision.to_loader[attribute][:location] == '::blank'
-              remove_copy_on_metis(revision, attribute)
-            end
+        revision.attribute_names.select {
+          |attribute| is_file_attribute(revision.model, attribute)
+        }.each do |attribute|
+          if !revision[attribute]  # nil
+            remove_copy_on_metis(revision, attribute)
+          elsif revision[attribute].start_with? 'metis:'
+            copy_file_on_metis(revision, attribute)
+          elsif revision[attribute] == '::blank'
+            remove_copy_on_metis(revision, attribute)
           end
         end
       end
@@ -125,7 +124,9 @@ class UpdateController < Magma::Controller
     #   metis://<project>/<bucket>/<folder path>/<file name>
     # Splitting the above produces
     #   ["metis", "", "<project>", "<bucket>", "<folder path>" ... "file name"]
-    metis_file_location_parts = revision.to_loader[attribute][:location].split('/')
+    metis_file_location_parts = revision[attribute].split('/')
+
+    # We need the filename here, so we call to_loader
     new_file_name = revision.to_loader[attribute][:filename]
 
     # At some point, when Metis supports changing project names,
--- a/lib/magma/server/update.rb
+++ b/lib/magma/server/update.rb
@@ -84,21 +84,21 @@ class UpdateController < Magma::Controller
       model_revisions.each do |revision|
 
         revision.attribute_names.select {
-          |attribute| is_file_attribute(revision.model, attribute)
-        }.each do |attribute|
-          if !revision[attribute]  # nil
-            remove_copy_on_metis(revision, attribute)
-          elsif revision[attribute].start_with? 'metis:'
-            copy_file_on_metis(revision, attribute)
-          elsif revision[attribute] == '::blank'
-            remove_copy_on_metis(revision, attribute)
+          |attribute_name| is_file_attribute(revision.model, attribute_name)
+        }.each do |attribute_name|
+          if !revision[attribute_name]  # nil
+            remove_copy_on_metis(revision, attribute_name)
+          elsif revision[attribute_name].start_with? 'metis:'
+            copy_file_on_metis(revision, attribute_name)
+          elsif revision[attribute_name] == '::blank'
+            remove_copy_on_metis(revision, attribute_name)
           end
         end
       end
     end
   end
 
-  def remove_copy_on_metis(revision, attribute)
+  def remove_copy_on_metis(revision, attribute_name)
     # First get the current value of the file attribute,
     #   so we can get the file extension.
     # When we need to construct the link filename
@@ -108,7 +108,7 @@ class UpdateController < Magma::Controller
     return
   end
 
-  def copy_file_on_metis(revision, attribute)
+  def copy_file_on_metis(revision, attribute_name)
     host = Magma.instance.config(:storage).fetch(:host)
 
     client = Etna::Client.new(
@@ -124,10 +124,10 @@ class UpdateController < Magma::Controller
     #   metis://<project>/<bucket>/<folder path>/<file name>
     # Splitting the above produces
     #   ["metis", "", "<project>", "<bucket>", "<folder path>" ... "file name"]
-    metis_file_location_parts = revision[attribute].split('/')
+    metis_file_location_parts = revision[attribute_name].split('/')
 
     # We need the filename here, so we call to_loader
-    new_file_name = revision.to_loader[attribute][:filename]
+    new_file_name = revision.to_loader[attribute_name][:filename]
 
     # At some point, when Metis supports changing project names,
     # this parameter should be the old file project name (metis_file_location_parts[2]))
--- a/lib/magma/server/update.rb
+++ b/lib/magma/server/update.rb
@@ -1,3 +1,4 @@
+require 'pry'
 require_relative 'controller'
 
 class UpdateController < Magma::Controller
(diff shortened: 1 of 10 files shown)


lib/magma/server/update.rb
```diff
@@ -55,4 +60,67 @@ def load_revisions
...
+
+          host = Magma.instance.config(:storage).fetch(:host)
+
+          client = Etna::Client.new(
+            "https://#{host}",
+            @request.cookies[Magma.instance.config(:token_name)] || auth.send('auth', *[@request, :etna]))
+
+          copy_route = ''
+
+          client.routes.each do |route|
```
coleshaw, line 92: Use Ruby "find" instead of iterating.

lib/magma/server/update.rb
```diff
@@ -55,4 +60,67 @@ def load_revisions
...
+            if route[:name] == 'copy'
+              copy_route = route
+              break
+            end
+          end
+
+          if copy_route == ''
```
coleshaw, line 99: next unless copy_route

lib/magma/server/update.rb
```diff
@@ -55,4 +60,67 @@ def load_revisions
...
+        if revision.to_loader.key?(:stats)
+          auth = Etna::Auth.new(:magma)
+
+          host = Magma.instance.config(:storage).fetch(:host)
+
+          client = Etna::Client.new(
+            "https://#{host}",
+            @request.cookies[Magma.instance.config(:token_name)] || auth.send('auth', *[@request, :etna]))
+
+          copy_route = ''
```
coleshaw, line 90: Don't need to initialize if using 'find' below

lib/magma/server/update.rb
```diff
@@ -55,4 +60,67 @@ def load_revisions
...
+            end
+          end
+
+          if copy_route == ''
+            next
+          end
+
+          path = client.send(
+            'route_path',
+            *[copy_route, {:project_name => @project_name, :bucket_name => "magma", :file_path => "test.txt"}])
```
coleshaw, line 105: Don't have hardcoded filename here?

coleshaw: Syntax is `{project_name: @project_name ...}`

lib/magma/server/update.rb
```diff
@@ -55,4 +60,67 @@ def load_revisions
...
+        # The below test for :stats as a File indicator seems
+        #   brittle -- anything better?
+        if revision.to_loader.key?(:stats)
+          auth = Etna::Auth.new(:magma)
+
+          host = Magma.instance.config(:storage).fetch(:host)
+
+          client = Etna::Client.new(
+            "https://#{host}",
+            @request.cookies[Magma.instance.config(:token_name)] || auth.send('auth', *[@request, :etna]))
```
coleshaw, line 88: Patch Etna::Auth to put the token into the request in `etna.user` in auth.rb, around line 80.
Then could do `@user.token`

lib/magma/server/update.rb
coleshaw: client.send('route_path', copy_route, project_name: etc., bucket_name: , file_path: blah). Don't need the `*[]`. By Ruby convention it knows the last options are in a hash.

coleshaw: Patch Etna::Client to expose `route_path` as public method.

lib/magma/server/update.rb
```diff
@@ -55,4 +60,67 @@ def load_revisions
...
+          # Now populate the standard headers
+          hmac_params = {
+            method: 'post',
+            host: host,
+            path: path,
+
+            expiration: (DateTime.now + 10).iso8601,
+            id: 'magma',
+            nonce: SecureRandom.hex,
+            headers: { project_name: @project_name, action: 'copy' },
```
coleshaw, line 116: Leave headers as empty Hash.

coleshaw: This is used to include extra parameters we want signed in HMAC signature, like user ID or something.

coleshaw: We will have optional parameters that should get hashed, `new_bucket_name`, `new_file_name`, and maybe `new_project_name`

lib/magma/server/update.rb
coleshaw: Bucket_name should be "old" / original bucket name of existing file.

lib/magma/server/update.rb
coleshaw: `new_bucket_name == magma`, `new_file_name` == magma's controlled name for the file. We should generate that somehow? From FileAttribute to see how it gets done now when a file gets put into S3.

coleshaw:
@graft , I believe this works. At least locally, when testing manually against Metis, I get a new entry in the database!

```
  4 | ipi          | test.txt                                 |         1 | f         | developer@ucsf.edu|Outis Developer | 2020-05-07 18:31:40.460931 | 2020-05-07 18:31:40.460931 |           |             6
  5 | ipi          | patient-IPIGYN001-flojo_file.txt         |         2 | f         | developer@ucsf.edu|Outis Developer | 2020-05-14 20:27:39.075861 | 2020-05-14 20:27:39.075861 |           |             6
(5 rows)
```

I'm missing the methods that handle `::nil` and `::blank` revisions, but want to put them into a separate PR, since this one seemed like it was getting big.

coleshaw:
@graft , I believe this works. At least locally, when testing manually against Metis, I get a new entry in the database!

```
  4 | proj          | test.txt                                 |         1 | f         | user | Developer | 2020-05-07 18:31:40.460931 | 2020-05-07 18:31:40.460931 |           |             6
  5 | proj          | model-record-attribute.txt         |         2 | f         | user | Developer | 2020-05-14 20:27:39.075861 | 2020-05-14 20:27:39.075861 |           |             6
(5 rows)
```

I'm missing the methods that handle `::nil` and `::blank` revisions, but want to put them into a separate PR, since this one seemed like it was getting big.

graft:
This is nice stuff. I might pay attention and simulate in the tests some sad-path stuff, like Metis replying with a 400 preventing the copy operation.

In this vein, if the user updates multiple records at once, you are sending several requests instead of a single large request. One or more of these requests could 400. Since we may have already asked metis to copy some of the entries successfully we have no choice but to link the record. On the other hand, we can't complete the entire update, since some of our requests have failed. This leaves us in a "partial update" state, very bad. I'm not really sure what to do in this case.

lib/magma/server/update.rb
```diff
@@ -83,16 +83,15 @@ def update_any_file_links
...
-            if !revision.to_loader[attribute]  # nil
-              remove_copy_on_metis(revision, attribute)
-            elsif revision.to_loader[attribute][:location].start_with? 'metis:'
-              copy_file_on_metis(revision, attribute)
-            elsif revision.to_loader[attribute][:location] == '::blank'
-              remove_copy_on_metis(revision, attribute)
-            end
+        revision.attribute_names.select {
+          |attribute| is_file_attribute(revision.model, attribute)
+        }.each do |attribute|
```
graft, line 88: seems trivial but I would use attribute_name here, it quickly becomes confusing whether you are dealing with the name or the actual object.

coleshaw:
Great points. I can add in more tests for non-successful copies.

Regarding multiple updates, I should add tests for those, too. What are your thoughts on using Threads or Fibers to wait for all requests to return, like [in this question](https://stackoverflow.com/questions/6643964/asynchronous-http-request-in-ruby)? If all requests return 200, we continue on to `load_revisions`. If one or more fail, then we clean up by deleting the successful ones on Metis.

graft:
The async nature of threads doesn't seem to buy anything here since we must still block, and the problem is rather that we might be partially successful. So far the validation logic tries to be transactional - either the whole update works or not.

coleshaw:
Hm, sorry, not quite sure I understand why something like this approach won't help avoid the partially successful problem. With Threads or some sort of async operation, if we block all the updates, can't we "undo" the successful copies on the Metis side if any single Metis copy fails? From the original point (just for context):

> In this vein, if the user updates multiple records at once, you are sending several requests instead of a single large request. One or more of these requests could 400. Since we may have already asked metis to copy some of the entries successfully we have no choice but to link the record **<Here, can't we "remove" the copy entries that succeeded on Metis, then throw an exception and skip the next steps on Magma? I'm confused about why we are left with no choice but to proceed with the linking on Magma?>**. On the other hand, we can't complete the entire update, since some of our requests have failed. This leaves us in a "partial update" state, very bad. I'm not really sure what to do in this case.

Maybe this is easier discussed verbally (and maybe with a whiteboard!)?

coleshaw:
Just a random question about the partial update problem ... is that something that you think we need to address now, in this PR, or something we should address later? You can't update multiple records through the new Timur UI right, so is that more of a future-issue we tackle later?

lib/magma/server/update.rb
```diff
@@ -55,4 +60,119 @@ def load_revisions
...
+    }
+
+    hmac = Etna::Hmac.new(Magma.instance, hmac_params)
+
+    # These get re-encoded into query string by Etna Client
+    hmac_params = Rack::Utils.parse_query(
+      hmac.url_params[:query]).map { |k,v| [k.to_sym, v] }.to_h
+
+    client.send(
+      'query_request',
```
graft, line 170: Can this be done with body_request? That method looks similar to `MetisClient#post`, which I use to post to hmac url + params with a post body.

lib/magma/server/update.rb
```diff
@@ -55,4 +60,119 @@ def load_revisions
...
+      Net::HTTP::Post,
+      hmac.url_params[:path],
+      hmac_params.merge(bulk_copy_params))
+  rescue Etna::Error => e
+    log(e.message)
+    @errors.concat([e.message])
```
graft, line 176: This, presumably, destroys `success?`, which then causes `load_revisions` not to run?

coleshaw:
Yes, any errors in `execute_bulk_copy_on_metis` will cause `load_revisions` to not run. [There is a spec](https://github.com/mountetna/magma/pull/139/files#diff-f999d89edde83b171d0e86e801229cdcR159) for that scenario.

coleshaw:
@graft , refactored with `body_request` per your feedback. Thoughts on current state? I'll branch off of this for the continuing work so this PR doesn't get too big.

coleshaw:
Superseded by #180

PR #140, opened 2020-05-14
Add attr validation options to attributes table

alimi:
This PR follows up on https://github.com/mountetna/magma/pull/135 by adding attribute validation options to the attributes table. Like the other `Magma::Attribute::EDITABLE_OPTIONS`, validation options stored in the attributes table will be loaded in favor of any defined on models in Ruby.

## Dealing with cached validation objects

There is a little weirdness around updating validation options that are already loaded in memory. When we convert validations to validation objects, we cache the validation object.

https://github.com/mountetna/magma/blob/453cf160ffc7207a4ac7e25d2bb027ebc683e179/lib/magma/attribute.rb#L29-L31

So if a validation that was already loaded in memory is changed in the attributes table, the validation object will still be based on the old validation data.

```ruby
irb(main):002:0> Ipi::Patient.attributes[:sop_version].validation_object
=> #<Magma::ArrayValidationObject:0x000055a9ec68c248 @object=["SOP1", "SOP2", "SOP3"]>
irb(main):008:0> Magma.instance.db[:attributes].insert(project_name: "ipi", model_name: "patient", attribute_name: "sop_version", validation: Sequel.pg_json_wrap({ type: "Array", value: ["SOP4", "SOP5"] }))
=> nil
irb(main):009:0> Ipi::Patient.attributes[:sop_version].validation_object
=> #<Magma::ArrayValidationObject:0x000055a9ec68c248 @object=["SOP1", "SOP2", "SOP3"]>
```

[`Magma::Attribute#update_option`](https://github.com/mountetna/magma/blob/88ddc3d4ba626edeb508d651afbe9c7387349c73/lib/magma/attribute.rb#L124) has been updated to reset the cached validation object so this might no be a huge issue. However, if we update validations some other way we'll have to watch out for this. Alternatively, we could stop caching the validation object but I'm not sure if that's a better approach.

## Migration version conflicts

This PR adds a migration, `db/migrations/002_add_validation_to_attributes.rb`, and https://github.com/mountetna/magma/pull/134 adds a migration, `db/migrations/002_add_link_model_to_attribute.rb`. After whichever PR gets merged, the other will have to change its migration version to 003. /cc @notmarkmiranda 

See https://github.com/mountetna/magma/issues/126.

### DIFF
--- a/db/migrations/002_add_validation_to_attributes.rb
+++ b/db/migrations/002_add_validation_to_attributes.rb
@@ -1,0 +1,7 @@
+Sequel.migration do
+  change do
+    alter_table(:attributes) do
+      add_column :validation, :json
+    end
+  end
+end
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -1,7 +1,7 @@
 class Magma
   class Attribute
     DISPLAY_ONLY = [:child, :collection]
-    EDITABLE_OPTIONS = [:description, :display_name, :format_hint]
+    EDITABLE_OPTIONS = [:description, :display_name, :format_hint, :validation]
 
     attr_reader :name, :loader, :validation, :format_hint, :unique, :index, :restricted
 
@@ -27,7 +27,7 @@ class Magma
     end
 
     def validation_object
-      @validation_object ||= Magma::ValidationObject.build(@validation)
+      @validation_object ||= Magma::ValidationObject.build(@validation&.symbolize_keys)
     end
 
     def json_template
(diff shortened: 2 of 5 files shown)


graft:
I think this looks good - I am a bit nervous about something like type: Regexp being built from a string in the JSON but defined as a /Regexp/ in the ruby models (or the same for Range, etc.). I think it should all work out but can we perhaps add to the validation_object tests to prove they can actually validate as they ought to?

graft:
Err, re: the caching issue, doesn't seem like a huge deal - going in through update_options seems like a nice way to solve the problem.

alimi:
Good call on updating the validation object specs! I added more coverage in https://github.com/mountetna/magma/pull/140/commits/c094c943c7d2b40ea35f94669ff2418d3e0a2aef.

graft:
:candy: :ship:

Issue #40, opened 2017-12-06
Unify update + loader endpoints


graft:
The loader endpoint (not yet built) uses the Magma::Loader interface to push records into a table. This interface has a lot of nice features: it computes validations, it can build associations across tables in bulk, and it uses a bulk insertion mechanism to update many records at once. Meanwhile the update endpoint uses the Sequel ORM. This is mostly because of Carrierwave, without which we cannot post files, and which attaches through Sequel. However, using the ORM for many records is very slow.

The Update endpoint should use the Magma::Loader interface to insert or update records. File uploads might have to be handled separately

graft:
There are several different document formats being defined here, none of which are very well-documented.

1. "revision" format, accepted by the /update endpoint, which is  scattered across the attribute framework in its definition. Essentially, here, there are no table insertions, you may insert links to non-existent records (thus allowing record creation through linking); links are made transparently using identifiers, e.g. { collection: [ 'ID1', 'ID2', 'ID3' ] }

2. "loader" format, whose definition is present mostly in update_entry, insert_entry and temp_entry, all of which define facilities for inserting data into the database. The input format is not well-specified, but links may still be made using identifiers.

3. "payload" format, returned by the "retrieve" endpoint. Similar to the "revision" format, but crucially the /update endpoint may represent a record differently (an update_record rather than a retrieve_record)

graft:
The 'revision' format has been enshrined in the 'Revision' object (which previously processed a set of "revisions", incorrectly). This clarifies a great deal in the update process.

However, much confusion remains. Currently, each attribute manages conversion between input formats and output formats. The input formats include Magma::Revision and the results from Magma::Question. The output formats include Magma::Payload (json), Magma::Payload (tsv) and Magma::Loader. This provides for four possible conversions, Magma::Revision => Magma::JsonPayload, Magma::Revision => Magma::Loader, Magma::Question => Magma::JsonPayload, Magma::Question => Magma::TsvPayload

Right now these conversions happen through five methods on the attribute, json_payload, text_payload, update & update_links, and update_payload. Previously these were not understood (by me) as format-conversion interfaces so they are somewhat poorly defined; they are essentially helper methods for the conversion done by Magma::Revision or Magma::Payload. Note that in the Question case, the output format is doing the conversion work, while in the Revision case, the revision is doing the conversion work.

[...]

PR #141, opened 2020-05-19
Skip update file spec

notmarkmiranda:
This will temporarily skip the update file spec until we have a chance to fix it.


graft:
:+1: this ought to fix whenever we close #40, but at this point we are planning on keeping that open until all of the magma/metis stuff is done.

notmarkmiranda:
👍🏼 sounds good.

graft:
Woops, sorry to be unclear - I meant to say: we should merge this, because #40 won't close for a long while and it is bad to have this failing test in there...

Issue #142, opened 2020-05-19
Fix link_model naming


graft:
A few errors crept in following #134:
https://github.com/mountetna/magma/blob/82bfd81ce816ca972ca320ccc8013842caec5bb3/lib/magma/attributes/child.rb#L14

https://github.com/mountetna/magma/blob/82bfd81ce816ca972ca320ccc8013842caec5bb3/lib/magma/project.rb#L26

Both of these instances should be amended to use `link_model`.

PR #144, opened 2020-05-19
Fix link_model naming

notmarkmiranda:
#142 

This will fix 2 places where `link_model_name` was changed incorrectly.


graft:
:ship:

PR #147, opened 2020-05-21
Clean up json_template

notmarkmiranda:
#128 

This removes model_name from json_template and adds the validation_object
directly to json_template.

### DIFF
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -49,6 +49,8 @@ class Magma
     def json_template
       {
         name: @name,
+        model_name: self.is_a?(Magma::Link) ? link_model.model_name : nil,
+        link_model_name: self.is_a?(Magma::Link) ? link_model.model_name : nil,
         type: database_type.respond_to?(:name) ? database_type.name : database_type,
         attribute_class: attribute_class_name,
         desc: description,
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -59,7 +59,8 @@ class Magma
         restricted: @restricted,
         format_hint: @format_hint,
         read_only: read_only?,
-        shown: shown?
+        shown: shown?,
+        validation_object: validation_object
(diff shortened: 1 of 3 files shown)


spec/magma_attribute_spec.rb
```diff
@@ -177,5 +177,14 @@
...
       expect(attribute.validation_object.validate("A")).to eq(true)
     end
+
+    it "contains a ValidationOjbect" do
+      model = double("model", project_name: :project, model_name: :model)
+      attribute = Magma::Attribute.new("name", model, {
+        validation: { type: "Regexp", value: /^[a-zA-Z]{1}$/ }
+      })
+
+      expect(attribute.json_template[:validation_object]).to be_a(Magma::RegexpValidationObject)
```
alimi, line 187: I think `json_template[:validation_object]` should return JSON so Timur can consume it. One option could be to define `to_json` on `Magma::ValidationObject`. Then, we could call `validation_object.to_json` in `#json_template`.

lib/magma/attribute.rb
```diff
@@ -49,7 +49,6 @@ def validation_object
     def json_template
       {
         name: @name,
-        model_name: self.is_a?(Magma::Link) ? link_model.model_name : nil,
```
coleshaw, line 52: Does Timur not rely on this field? Sorry, I'm fairly ignorant of what fields Timur relies on, so just checking...

notmarkmiranda: I think you're right. I'll check Timur and make a corresponding PR if necessary before merging this one.

alimi: We've generally been trying to make non-breaking changes in Magma so Timur will keep working. So if Timur depends on `model_name`, we should keep it here.

graft: We should add a replacement field (link_model_name) but keep this in place until we patch timur

Issue #145, opened 2020-05-20
Move test Labors project attributes to the attributes table


alimi:
Now that the attributes table is complete (https://github.com/mountetna/magma/pull/132), we can move the Labors project attributes into the attributes table and remove the Ruby definitions.

PR #148, opened 2020-05-21
Move spec Labors model attributes to attributes table

alimi:
Before `spec/spec_helper.rb` starts the Magma server, it will load Labors model attributes from a CSV into the attributes table. Then when the server starts, it will be able to load the attributes in memory from the database.

If the CSV data is bad for any reason, we'll truncate the attributes table before raising an exception. This will leave the test database in a good state so we can try loading the data again.

Fixes https://github.com/mountetna/magma/issues/145


spec/fixtures/labors_model_attributes.csv
```diff
@@ -0,0 +1,33 @@
...
+labors,codex,tome,string,,,
+labors,codex,lore,match,,,
+labors,labor,project,parent,,,
+labors,labor,name,identifier,,,
+labors,labor,monster,child,,,
+labors,labor,number,integer,,,
+labors,labor,completed,boolean,,,
+labors,labor,year,date_time,,,
+labors,labor,prize,table,,,
+labors,labor,contributions,matrix,,,"{""type"":""Array"", ""value"":[""Athens"",""Sparta"",""Sidon"",""Thebes""]}"
```
alimi, line 18: Postgres uses `"` as its escape string for CSVs so we end up with lots of `""` in the validation JSON.

spec/spec_helper.rb
```diff
@@ -22,12 +22,33 @@
...
+
+def truncate_attributes_table_on_error(&block)
+  yield
+rescue => error
+  Magma.instance.db[:attributes].truncate
+  raise error
+end
+
+truncate_attributes_table_on_error do
+  Magma.instance.db.run(<<~SQL
```
graft, line 35: This is interesting - given that we might *often* want to do such a thing, even in production - should this become a command?

spec/fixtures/labors_model_attributes.csv
graft: What motivated the choice of CSV instead of JSON or YAML? the latter might be less compact in terms of number of lines but I think they would be more legible/editable and the CSV is sparsely populated so a YAML might not be so noisy.

spec/spec_helper.rb
```diff
@@ -22,12 +22,33 @@
 
 Magma.instance.configure(YAML.load(File.read('config.yml')))
 
+Magma.instance.setup_db
+
+def truncate_attributes_table_on_error(&block)
+  yield
```
graft, line 28: Can we assign DatabaseCleaner to do this for us?

spec/fixtures/labors_model_attributes.csv
alimi: I went with CSV because it feeds straight into Postgres' `COPY` command. This lets us take advantage of Postgres' features for importing data, which means less code that we have to maintain in the future. The `COPY` command is also pretty fast so we shouldn't have to worry about the import slowing down the overall test run time.

That being said, maintaining the CSV could also be challenging because it is less legible. Lemme play around with a YAML file and we can compare the two.

spec/spec_helper.rb
alimi: Yeah, that's an option. I guess the command would take a file path as an argument?

spec/fixtures/labors_model_attributes.csv
alimi: Ok, I switched it to YAML in https://github.com/mountetna/magma/pull/148/commits/3b5cc2a905de003ce6b6213074fdcce0d8b21da4. If we don't like it, I can undo the commit.

The YAML file is more human readable, but we have to do more work parsing it on import.

https://github.com/mountetna/magma/blob/3b5cc2a905de003ce6b6213074fdcce0d8b21da4/spec/spec_helper.rb#L35-L45

vs.

https://github.com/mountetna/magma/blob/a034bbfddc87fa1edeb6ad745f8fc8ac11ff28cd/spec/spec_helper.rb#L35-L40

It's a bummer we can't leverage Postgres' `COPY` command with the YAML file, but if the YAML is easier to read/edit it might be a worthwhile trade off. What do you think?

spec/spec_helper.rb
alimi: Maybe? Since we're loading the data before DatabaseCleaner is setup, we'd have to configure and start DatabaseCleaner earlier. I'm not sure it's worth the trouble given we're only truncating a single table.

spec/fixtures/labors_model_attributes.csv
graft: If it helps there is `Sequel::Model.multi_insert` which you can feed a bunch of hashes & bulk insert (the hashes all have to have the same keys). This uses Postgres' `INSERT`; `COPY` is more efficient, but in this case it is probably not that significant a difference - though we might want to benchmark the difference with ~10k items at some point.

You could also continue to use `COPY` by packing the YAML data into a CSV using `CSV.generate`.

alimi: Yeah, we're not inserting enough stuff for `COPY` to make a big difference.

Maybe we stick with what we have now. Seems like the best compromise 😄

alimi:
@graft is this PR good to merge?

graft:
:shipit:

PR #151, opened 2020-05-27
Use GitHub Actions to run test suite

alimi:
This PR sets up a GitHub Actions workflow that will run the test suite on every push and PR.

Here are a couple of example runs:
  * https://github.com/mountetna/magma/actions/runs/116143494
  * https://github.com/mountetna/magma/actions/runs/116903644

If you click on the "build" link on the left sidebar, you can see all the steps that happened for each run.

GitHub Actions also gives us a status badge that I've included in the [README](https://github.com/mountetna/magma/blob/c8bc242ad88e6892397aa175eaa782893398bef8/README.md).

## Magma changes

There are a few changes we had to make so GitHub Actions could run the test suite.

* Relax the Ruby version in the Gemfile because GitHub Actions only supports major.minor Ruby versioning. https://github.com/mountetna/magma/commit/100aa63cef4971741d6a4b2772ba6a0cd6b2b8fd
* Gemfile updates (https://github.com/mountetna/magma/commit/0ebad69c144c70e2222aaf68411cac3040416a0b)
  * Use the publicly released version of etna
  * Use the same version of sequel in both `Gemfile` and `Gemfile.lock`. Bundler uses the `Gemfile.lock` version (5.28.0).
* Add a `config.yml.test`. The workflow will copy that file to `config.yml` so it can boot magma.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,22 +15,34 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+    env:
+      MAGMA_ENV: test
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
     - name: Set up Ruby 2.5
       uses: actions/setup-ruby@v1
       with:
         ruby-version: 2.5
-    - name: Setup deps and run RSpec test suite
+    - name: Bundle install
+      run: |
+        gem install bundler -v 2.1.4
+        bundle config path vendor/bundle
+        bundle install --jobs 4 --retry 3
+    - name: Set up database
       env:
         PGHOST: 127.0.0.1
         PGUSER: developer
-        MAGMA_ENV: test
       run: |
-        gem install bundler -v 2.1.4
-        bundle install --jobs 4 --retry 3
         cp config.yml.test config.yml
         bin/magma create_db labors
         bin/magma migrate
         bin/magma global_migrate
+    - name: Run RSpec test suite
+      run: |
         rspec spec/
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,6 +48,6 @@ jobs:
       run: |
         bundle exec rspec spec/
     - name: Run test suite again to detect flaky failures
-      if: ${{ steps.first_test_run.outcome == "success" }}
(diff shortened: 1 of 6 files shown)


coleshaw:
This is great! Love having tests run for every push and PR. Kind of concerning that one check passes (for the push) while the other fails (for the PR), though ...  Both of the rspec logs show an `ERROR` around line 54, but one test run counts that as a pass, while the other counts it as a fail. It would be nice to figure out why, if there's something in the environment that needs to be set or changed?

alimi:
Yeah, there are a few flaky tests floating around. Lemme see if I can figure out what's going on with the one that failed on the PR build.

graft:
This is awesome, even the flaky tests, which it's nice to get some visibility on. We might want to turn the volume down on the spec output, I think we could set the log_level in config.yml.test to be `error` rather than `warn` to shut up all those log lines.

Couple of questions:

1) why are there two test runs, one for PR and one for Push?
2) Does the status badge reflect the most recent push to master or does it just reflect whatever branch was run most recently?

alimi:
>     1. why are there two test runs, one for PR and one for Push?

Yep, one was triggered by opening the PR and the other was triggered by pushing my last commit.

It doesn't show up in the PR checks, but there are also builds for every commit I pushed up. If you look through my commits, you'll see a green check for commits that had green builds and red x's for commits that had bad builds.

>     2. Does the status badge reflect the most recent push to master or does it just reflect whatever branch was run most recently?

I thought it was based of the last push to master, but it looks like it's using the last build. I think it makes more sense to base it off of the last push to master. Lemme see if we can change it.

alimi:
I've made a few changes. First, the flaky test has been fixed in https://github.com/mountetna/magma/pull/151/commits/fd9d8c3cb60c901503118f69578cc690cacadb33.

https://github.com/mountetna/magma/pull/151/commits/21eb90d052481f72e23e5fdbbea1a5da5b1576d6 changes the log levels so we won't see so much noise in the test run output.

Finally per the conversation we had in Slack yesterday, we'll run the test suite twice to try to catch any flaky tests that creep into the test suite. We'll only do the second run if the first run passes because if the first run fails we'll already know if there's something we need to fix. So far the double run hasn't caught any flaky tests 🤞🏾.

Also, there is a small difference between the two runs that get triggered by a PR and push. The push run is based off of the last pushed commit, while the PR run is based on a potential merge commit of the PR into master. So the PR run gives us a little more confidence that the changes will work with whatever is on master. h/t @notmarkmiranda

alimi:
The status badge is based off of master by default so nothing we need to change there. https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

graft:
:ship:

PR #152, opened 2020-05-29
Validate attributes exist in model tables

### LINKED ISSUE #130: Check attributes at boot time
Currently in `Magma::Project#load_project`, Magma checks that there is a table for the model. It should also check that there is a column for each attribute in the attribute table. If there is not, magma should raise a validation error and report the discrepancy.

alimi:
To help make `Magma#validate_models` a little more digestible, I also moved Magma attribute validations into `#validate_attributes`.

Fixes #130

### DIFF
--- a/lib/magma/attributes/link.rb
+++ b/lib/magma/attributes/link.rb
@@ -35,5 +35,9 @@ class Magma
     def link_identity
       link_model.attributes[link_model.identity]
     end
+
+    def missing_column?
+      false
+    end
   end
 end
--- a/lib/magma/attributes/link.rb
+++ b/lib/magma/attributes/link.rb
@@ -35,9 +35,5 @@ class Magma
     def link_identity
       link_model.attributes[link_model.identity]
     end
-
-    def missing_column?
-      false
-    end
   end
 end
--- a/spec/magma_spec.rb
+++ b/spec/magma_spec.rb
@@ -1,10 +1,8 @@
 require_relative '../lib/magma'
 require 'yaml'
 
-def disconnect_attribute model, att_name
-  att = model.attributes[att_name]
-  model.attributes[att_name] = nil
-  att
+def disconnect_attribute(model, att_name)
+  model.attributes.delete(att_name)
 end
 
 def reconnect_attribute model, att_name, value
@@ -41,6 +39,7 @@ describe Magma do
         expect{Magma.instance.validate_models}.to raise_error(Magma::ValidationError)
       end
     end
+
     context 'project model' do
       before(:each) do
         # Delete the constant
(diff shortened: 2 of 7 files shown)


lib/magma/attributes/link.rb
```diff
@@ -35,5 +35,9 @@ def link_records(identifiers)
     def link_identity
       link_model.attributes[link_model.identity]
     end
+
+    def missing_column?
+      false
+    end
```
alimi, line 41: Since link attributes like `Magma::ChildAttribute` don't have columns on their model's tables, they'll always return false for `#missing_column?`.

spec/magma_spec.rb
```diff
@@ -1,10 +1,8 @@
 require_relative '../lib/magma'
 require 'yaml'
 
-def disconnect_attribute model, att_name
-  att = model.attributes[att_name]
-  model.attributes[att_name] = nil
-  att
+def disconnect_attribute(model, att_name)
+  model.attributes.delete(att_name)
```
alimi, line 5: In production, it's pretty unlikely we'd define nil attributes on a model. So instead of leaving nil attributes behind in the tests, we can completely remove them.

lib/magma/attributes/link.rb
graft: Magma::ForeignKeyAttributes are also link_attributes, meaning `parent` and `link` types require a column.

graft:
Not relevant to this PR, but I just noticed the link validation is a bit bizarre and doesn't seem correct, or it's at least very provisional and not rigorous enough. For example, you could have two models with "collections" of each other - each expects the other to hold a foreign key, but neither does because there is no join model - but they should check out as reciprocally linked. Also, a self-referential link (which is in use, e.g., in the `ipi::patient` model for setting a `reference_patient`) checks out - but only because it proves to be its own reciprocal! Bizarre and strange. These behaviors should be clarified before we start trying to figure out how to make mutations, since in the future we'll have to validate graph integrity at every step.

lib/magma/attributes/link.rb
alimi: Gotcha. This should be fixed in https://github.com/mountetna/magma/pull/152/commits/353ab82de2895033c0f2d88f08c4934bf4bd9b93.

alimi:
I added a new issue to clean up the link validation. https://github.com/mountetna/magma/issues/160

alimi:
The flaky tests (https://github.com/mountetna/magma/pull/152/checks?check_run_id=728417245) are fixed in https://github.com/mountetna/magma/pull/155.

alimi:
I rebased against master to pull in the flaky test fixes. I think this should be good to merge.

graft:
It seems a fine PR, :shipit:

PR #155, opened 2020-05-29
Update retrieve TSV spec

alimi:
This fixes a flaky failure that happened after https://github.com/mountetna/magma/pull/151.

https://github.com/mountetna/magma/runs/721348936

I'm assuming the TSV header order doesn't matter, so I used the `match_array` helper to verify the header elements are present without worrying about the order. If the order matters, we can try something else.

### DIFF
--- a/lib/magma/retrieval.rb
+++ b/lib/magma/retrieval.rb
@@ -32,6 +32,16 @@ class Magma
           !(att.is_a?(Magma::TableAttribute) && @collapse_tables)
         end
 
+        if @requested_attribute_names != "all"
+          attributes.sort_by! do |att|
+            if att.name == @model.identity
+              -1
+            else
+              @requested_attribute_names.index(att.name)
+            end
+          end
+        end
+
         # if there is no identifier, use the :id column
         if !@model.has_identifier?
           attributes.push(OpenStruct.new(name: :id))
--- a/spec/retrieve_spec.rb
+++ b/spec/retrieve_spec.rb
@@ -136,8 +136,8 @@ describe RetrieveController do
       json = json_body
 
       # any model with an identifier returns all records
-      expect(json[:models][:labor][:documents].keys).to match(labors.map(&:name).map(&:to_sym))
-      expect(json[:models][:monster][:documents].keys).to match(monsters.map(&:name).map(&:to_sym))
+      expect(json[:models][:labor][:documents].keys).to match_array(labors.map(&:name).map(&:to_sym))
+      expect(json[:models][:monster][:documents].keys).to match_array(monsters.map(&:name).map(&:to_sym))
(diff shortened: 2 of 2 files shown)


graft:
In this case there are a bunch of things going on. The request to /retrieve includes attribute_names, which is an array of names, i.e. there is an ordering specified. In a TSV i'd expect the columns to come back in that order. Instead magma seems to just return in the order of model.attributes (`lib/magma/retrieval.rb:29`). The model ordering should be in the order they were loaded.

However, i think now loading from the db, that ordering has become unspecified. Previously the test coincidentally requested attribute names in the right order. If it had requested, and expected, in an order other than the model-order, it would have been failing already.

Under these circumstances I think the correct way to fix the test would be to sort the attributes in `Magma::Retrieval` correctly. It would also be nice to have a consistent ordering of model.attributes, but I don't think that's the issue here.

graft:
We should merge this, but make an issue for consistent attribute ordering in retrieve.

alimi:
Gotcha. I gave ordering `Magma::Retrieval#attributes` a shot in https://github.com/mountetna/magma/pull/155/commits/3bed1265aeb1ca80df5f6dddd36a88eba5b9fed9.

alimi:
We have another unrelated ordering failure:

https://github.com/mountetna/magma/blob/c93811f7eeb106672a83edbc1aa5d2b9315642c8/spec/retrieve_spec.rb#L237

```
   1) RetrieveController tables retrieves table associations
     Failure/Error: expect(models[:labor][:documents][:'Lernean Hydra'][:prize]).to eq(hydra_prizes.map(&:id))

       expected: [111, 112, 113]
            got: [113, 111, 112]
```

https://github.com/mountetna/magma/pull/155/checks?check_run_id=727672362

Does the ordering matter here @graft?

graft:
:shrug: on how i'm supposed to reply to these conversations in-line - err, that new spec doesn't seem to matter? That is, if it is just checking the keys of a hash I can't expect that to preserve ordering in the way that a TSV column header would.

alimi:
Heh, yeah it's tricky responding to comments. You can _Quote reply_ to a comment. It's in the ... menu.

image

> that new spec doesn't seem to matter? That is, if it is just checking the keys of a hash I can't expect that to preserve ordering in the way that a TSV column header would.

👍🏾 I'll update the assertion.

alimi:
Ok, https://github.com/mountetna/magma/pull/155/commits/77aeb61c1d839ee9d48dae40acc117415b6f9a06 looks to have fixed more flakiness in `spec/retrieve_spec.rb`.

graft:
Looks good to merge, let's :ship: so our tests aren't flaky!

PR #156, opened 2020-05-29
Fix root route

graft:
Trying to view magma in the browser should give you a basic status message. Instead, you get a 503 because the rack format for the route '/' is broken! This PR repairs this issue.


coleshaw:
This works for me locally, but it appears "bad" because of the flaky test being addressed in #155 . So I'd say it looks good to me, since the other issue is being addressed.

PR #157, opened 2020-06-01
Add load_project command

notmarkmiranda:
This adds a command `bin/magma load_project path/to/file.json`. It
parses the stored JSON and add attributes to the attributes table.

I'm encountering a few issues:
- knowing how to retrieve the `project_name` that is associated
- the `project` model does not have a `type` included

However, these may be due to the JSON I chose to store in the JSON.


lib/magma/commands.rb
```diff
@@ -2,6 +2,42 @@
...
+        attributes.keys.each do |att|
+          attribute = attributes[att]
+
+          db[:attributes].insert(
+            project_name: 'i dont know where to find the project name',
+            model_name: model_name,
+            attribute_name: attribute["name"],
+            type: attribute["type"] || "project",
+            display_name: attribute["display_name"],
+            # attribute["shown"] is the opposite of hidden?
```
graft, line 27: We should resolve this in json_template and use "hidden", "shown" was always a terrible idea.

lib/magma/commands.rb
```diff
@@ -2,6 +2,42 @@
...
+            hidden: attribute["shown"],
+            validation: attribute["validation_object"],
```
graft, line 29: If you rename `validation_object` to `validation`, alias `name` to `attribute_name`, and `shown` to `hidden`, you could just paste in the attribute directly here.

lib/magma/commands.rb
```diff
@@ -2,6 +2,42 @@
...
+
+      models.keys.each do |model|
+        model_name = models[model]["template"]["name"]
+        attributes = models[model]["template"]["attributes"]
+
+        attributes.keys.each do |att|
+          attribute = attributes[att]
+
+          db[:attributes].insert(
+            project_name: 'i dont know where to find the project name',
```
graft, line 22: Hmm - doesn't return in the payload from /retrieve at the moment, so I guess it would be the first argument to the `load_project` method?

lib/magma/commands.rb
```diff
@@ -2,6 +2,42 @@
...
+            attribute_name: attribute["name"],
+            type: attribute["type"] || "project",
+            display_name: attribute["display_name"],
+            # attribute["shown"] is the opposite of hidden?
+            hidden: attribute["shown"],
+            validation: attribute["validation_object"],
+            restricted: attribute["restricted"],
+            read_only: attribute["read_only"],
+            link_model_name: attribute["link_model_name"]
+            # where does attribute_class go?
```
alimi, line 33: `attribute_class` is only used by Timur. I think it's mostly used to decide how different attributes should be rendered. It doesn't need to be inserted into the `attributes` table.

alimi:
>     * the `project` model does not have a `type` included

Do other models have types? I thought type was only set on attributes.

lib/magma/commands.rb
graft: We'll deprecate it shortly anyway, since the `attribute_type` (which should be in here) should replace it.

lib/magma/commands.rb
```diff
@@ -2,6 +2,42 @@
...
+        attributes = models[model]["template"]["attributes"]
+
+        attributes.keys.each do |att|
+          attribute = attributes[att]
+
+          db[:attributes].insert(
+            project_name: 'i dont know where to find the project name',
+            model_name: model_name,
+            attribute_name: attribute["name"],
+            type: attribute["type"] || "project",
```
graft, line 25: not sure I understand the `|| "project"` here - there shouldn't be a default for "type"?

notmarkmiranda: This was a placeholder. In the JSON I retrieved for this test, the `project` doesn't have a `type`: 
image

spec/magma_commands_spec.rb
```diff
@@ -0,0 +1,300 @@
+describe 'Magma Commands' do
+  describe Magma::LoadProject do
+    let(:json_file) { './spec/fixtures/template.json' }
+
+    subject(:load_project) { described_class.new.execute('project_name', json_file) }
+
+    let(:attributes) { Magma.instance.db[:attributes] }
+
+    it 'loads attributes into the database' do
+      json_attributes = JSON.parse(JSON.parse(File.open(json_file).read))["models"]["aspect"]["template"]["attributes"]
```
graft, line 10: I find this double-parse confusing, can you clarify what this is doing?

spec/magma_attribute_spec.rb
```diff
@@ -183,7 +183,7 @@
       attribute = Magma::Attribute.new("name", model, {
         validation: { type: "Regexp", value: /^[a-zA-Z]{1}$/ }
       })
-      json_validation_object = attribute.json_template[:validation_object].as_json
+      json_validation_object = attribute.json_template[:validation]
```
graft, line 186: Despite its clearly inapt name, `#json_template` doesn't actually return json, it just returns an object that can be successfully turned into json - the actual transformation to json happens very late, [here](https://github.com/mountetna/magma/blob/40601560edc2d97dd6ef18880acad21c513e789a/lib/magma/server/retrieve.rb#L117) after the json_template is bundled with a bunch of others - so I think this `attribute.json_template[:validation]` should not yet be a JSON string?

spec/magma_commands_spec.rb
notmarkmiranda: I meant to point this out as a weird thing that is happening. Unsure of why I need to double parse, was hoping someone had knowledge of it than I would? Maybe the file being stored in the fixtures is written incorrectly?

graft:
Fantastic work on all of the command specs, it's great to have that additional coverage.

spec/magma_attribute_spec.rb
notmarkmiranda: That's what I originally had, but it did not get converted to a JSON string at any point before getting called by the retrieve endpoint.

graft:
Okay a few things:

First, I note that your fixtures file is a JSON string, that is, it is double-encoded JSON. This requires double-parsing, as you do in the spec and in `commands.rb`. You should certainly resolve this; I also might suggest `JSON.pretty_generate` for legibility.

I am noting this: https://github.com/mountetna/magma/blob/58f04455a26964c451866da003683d56bc95eba1/lib/magma/validation_object.rb#L28

Usually as_json would return a hash, and then a subsequent to_json can call as_json to render a real hash. The result of this is that when you call `@payload.to_hash.to_json`, #to_json will call `#as_json` on the validation_object, which will return a JSON string. #to_json will then call #to_json on this string, which will double-encode it.

Thus, the "validation" lines in your fixtures file are actually *triple* encoded JSON. Since you only decode twice (which works for the rest of the fixture), this means that the value "null" (which is the string `"null"`, not the value `null`, which is valid JSON) remains encoded. This actually ends up working out okay, because the Sequel::Dataset#insert seems to expect a JSON-encoded string in this spot, so when you insert, it successfully makes it into the database as `null`, rather than `"null"`, and so comes out as `nil` - then `#as_json` turns it back to the string "null" in the encoded output, and the test passes.

I'll note that the `Aspect` model only contains null validations, and you might be better off testing the `Monster` model, which actually contains a validation, and see if you can load/unload that successfully. Better yet, I think you ought to use this command (or its facilities) to replace Ali's loading from a fixture yaml in `spec/spec_helper.rb`, which will implicitly exercise the load_project operation in order to allow the remaining tests to pass.

notmarkmiranda:
I've fixed the double / triple encoded JSON and removed the `as_json` call to the ValidationObject, but now the call to `model#json_template` returns `"validation"=>"#<Magma::NullValidationObject:0x00007ffbb83766b0>"`.

spec/magma_attribute_spec.rb
```diff
@@ -183,9 +183,9 @@
...
 
-      expect(json_validation_object).to eq("{\"type\":\"Regexp\",\"value\":\"(?-mix:^[a-zA-Z]{1}$)\"}")
+      expect(json_validation_object).to be_a(Magma::RegexpValidationObject)
```
alimi, line 188: This spec is in the `describe "#validation_object"` block, but it should be in the `describe "#json_template"` block at the beginning of the file.

Now that the `validation_object` is included as a hash in the `json_template`, this spec needs to be updated.

```
  1) Magma::Attribute#validation_object contains a ValidationOjbect
     Failure/Error: expect(json_validation_object).to be_a(Magma::RegexpValidationObject)
       expected {:type=>"Regexp", :value=>/^[a-zA-Z]{1}$/} to be a kind of Magma::RegexpValidationObject
     # ./spec/magma_attribute_spec.rb:188:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:185:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:184:in `block (2 levels) in <top (required)>'
```

spec/fixtures/template.json
```diff
@@ -0,0 +1,152 @@
+{
+  "models": {
+    "monster": {
+      "documents": {
+        "monster7": {
```
graft, line 5: This "documents" section is unnecessary and could be axed.

lib/magma/validation_object.rb
```diff
@@ -24,8 +24,8 @@ def options
     def match
     end
 
-    def as_json
-      @options.to_json
+    def to_hash
+      @options&.to_hash
     end
```
graft, line 29: The usual behavior of #to_json on generic_object is to call as_json.to_json - so in line 65 of `attribute.rb` you could write `validation: validation_object`. Later, when the retrieve endpoint calls to_json on the payload, to_json will call `attribute[:validation].as_json.to_json` - so here you could write:

    def as_json
      @options
    end

`@options` should already be a Hash and will respond correctly to `Hash#to_json` when called. The result will be the same as you've written but this method seems to be the idiomatic intent of `#as_json`.

lib/magma/commands.rb
```diff
@@ -2,6 +2,50 @@
 require 'logger'
 
 class Magma
+  class LoadProject < Etna::Command
+    usage '[project_name, path/to/file.json] # Import attributes into database for given project name from JSON file'
+
+    ATTRIBUTES_OPTIONS = [
+      'attribute_name',
```
graft, line 9: Seems brittle to list these here - could we use `Magma::Attribute.options` instead + a few stragglers in a merge?

lib/magma/commands.rb
```diff
@@ -2,6 +2,50 @@
...
+      'read_only',
+      'unique',
+      'index',
+      'restricted'
+    ]
+
+    def execute(project_name, file_name)
+      file = File.open(file_name)
+      file_data = JSON.parse(file.read)
+
```
graft, line 27: I like: `JSON.parse(File.read(file_name), symbolize_names: true)` - which also saves you from using strings below.

lib/magma/validation_object.rb
```diff
@@ -24,8 +24,8 @@ def options
     def match
     end
 
-    def as_json
-      @options.to_json
+    def to_json
+      @options
```
notmarkmiranda, line 28: @graft Calling `.to_json` here would convert it to a JSON string before having it converted again by the retrieve endpoint. I left it as a hash to prevent needing to double parse the validation_object#options.

graft: The expected behavior of to_json is to return a JSON string. So you should have `@options.to_json` here and not call to_json in `#json_template`. In the test you might call `JSON.parse(attribute.json_template.to_json)` to test the output.

notmarkmiranda:
I made changes to converting the `validation_object`, the trade off is that `json_template` will be `null` for all attributes because within `json_template` in `attribute.rb` the object is not considered nil and deleted.

lib/magma/commands.rb
```diff
@@ -2,6 +2,38 @@
...
+      models = file_data[:models]
+
+      models.keys.each do |model|
+        model_name = models[model][:template][:name]
+        attributes = models[model][:template][:attributes]
+        attributes.values.each do |attribute|
+          attribute_type = attribute[:attribute_type]
+          attribute.slice!(*options)
+          attribute.merge!(
+            project_name: project_name,
```
graft, line 26: We don't do any validation of the input JSON in this command, so we must assume things are well-formed. Perhaps this should change at some point, but at this point we might want to at least ensure that `attribute[:attribute_name]` exists by putting it in the merge - we should be able to guarantee it if we iterate as `attributes.each do |attribute_name, attribute|`

graft:
The presence of `validation: null` in the template is annoying, but i'm not sure what else it is, i'm not really sure what the utility of compacting the attribute json is. We could probably force it out but let's just live with it.

I made one final comment about ensuring `attribute_name`, but let's :ship: after that.

Some future work:
1. Replace Ali's YAML fixture with a JSON one for loading spec/labors models in spec_helper.rb
2. When #162 closes this command ought to load models also.

lib/magma/commands.rb
notmarkmiranda: 👍🏼

Issue #143, opened 2020-05-19
Change file attributes to store JSON structure instead of just a string


coleshaw:
This will let us keep the original file name as well as the Metis link file path. Also allows us to store more metadata in Magma in the future.

Issue #51, opened 2017-12-06
Add metis support


graft:
Magma should use metis as its file service component.

Roadmapping:
Moved to [planning](https://github.com/mountetna/planning/issues/30)

graft:
The main reformations that must happen:

1) Magma does not accept File posts to /update. Instead, it returns a Metis upload-authorized URL. The client does whatever it wants with this URL and is responsible for uploading the file.

2) Metis notifies Magma that there is a file in the appropriate place via a notification endpoint, which it can reach via HMAC. Magma fills in the matching record.

graft:
To use Metis, magma should be able to do a few things:

1) Hold a path on metis in a file or image attribute. This uses the /:project_name/magma bucket on metis owned by user 'magma'. This path allows magma to generate get and put URLs that can be handed off to someone else to download or upload the file at that path. This doesn't require any new functionality from Metis or magma.

2) Move data from Metis into that path. Instead of providing a file to the /update endpoint, we send an existing Metis file path. Magma performs a move operation metis, transferring the file from wherever it is to Magma's preferred destination. Metis currently doesn't allow moves across buckets, so perhaps Metis needs a new command: /transfer, which lets you move across buckets.

How should we authenticate? We need to use the user's token in (2) because they must have the rights to access the file, as determined by Metis based on the bucket permissions and so on. However, Metis also needs to know that Magma approves of the transfer, so that users can't overwrite files using /transfer on Metis without Magma's approval. It would be nice to achieve both of these things in a single transfer.

We can do this if we have an endpoint that requires BOTH a valid token AND an Hmac-signed URL. With such an endpoint, we can *transfer* a file from bucket1/path1, where bucket1.owner = 'metis', to magma/path2, where magma.owner = 'magma'. So: the user requests Magma to upload from a specified Metis location. Magma uses the user's token, signs a url, and requests from Metis the transfer operation.

graft:
Using the transfer mechanism above we can avoid the necessity of requiring upload notification from Metis to Magma. The Magma upload cycle can instead look like this:

1) Post to /update with the revision "upload" for a file or image attribute. Magma replies with an upload URL for /:project_name/magma/uploads/<tmp_file_name> based on the model, attribute etc.

2) The user puts the file at the intended location.

3) When complete, the user posts to /update with the revision "metis://:project_name/magma/uploads/<tmp_file_name>". Magma moves the file using the Metis transfer endpoint and updates the DB when successful.

OR

[...]

PR #159, opened 2020-06-01
Revisions with ::temp return a temporary Metis path

coleshaw:
@graft , I think this is all that is needed for the temporary location?

One thought about the workflow ... this is reliant on storing the original filename in the database (requires JSON, as per #143), since we don't have a file extension associated with the temporary location. Unless we want Timur to also send the original filename for the `::temp` request to append the extension, but I think storing the JSON structure would be cleaner.


graft:
Hmm, you're right - there is no way to record the original filename in this workflow, somewhat of a problem. Not sure how to resolve that.

As for the temp link - I don't think a metis path suffices, because the user won't be allowed to upload into Magma's bucket. They could try to request authorization to write there, but since the bucket is owned by magma Metis ought to refuse. I think, instead, Magma must give the user a signed upload url to the temp path.

corps:
@coleshaw @graft 
Not sure about the workflow; metis calls into magma to get a temporary url, then I assume some client (server or browser) performs the actual upload right?
If it's the latter (the browser), the default behavior of forms submitting file inputs is to include the content type, which I believe s3 records.  Possibly the extension could be inferred from the recorded content type?

graft:
The workflow is described in #51, maybe? Anyway the gist:
1. The client has the original file and wishes to link it to a file attribute on a record in Magma.
2. The client asks Magma for a temporary file location on Metis to store the file.
3. Magma replies with a signed upload url to use on Metis.
4. The user uses the upload url to send the file, in chunks, to Metis at the temporary file location specified by Magma.
5. The user tells Magma to link the file at the temporary location to the record/attribute.
6. Magma tells Metis to copy the file at the temporary location to a permanent location.
7. Metis copies the files as specified and reports back to Magma
8. Magma stores `{ original_filename, metis_path }` in the file attribute for the given record.

So, in this act of prestidigitation, the question is: how does Magma get the original filename?

In the beginning, the user has the filename; Magma needs to end up with it. The user transmits three instructions: one to Magma, one to Metis, and one again to Magma. Therefore, either the user must communicate the filename directly to Magma in the first and third steps, or they must communicate it to Metis, which then communicates it to Magma.

Of this latter possibility, if the user could communicate original_filename to Metis, what would Metis do with it? There is no file metadata it can place this information into, only the upload filepath itself; this is given to the user by Magma. Hence, there is no space for Metis to preserve the information to communicate anything back to Magma.

There are two possibilities left:

1. The user sends the original_filename as part of the initial temp filename request. Magma cannot write this down in the database (yet), but it CAN encode it into the temp filename, e.g. "metis://athena/magma/temp-#{uid}-#{hex_encode(file_name)}". Later, when the user links the temp file back into the record, Magma recognizes its own encoding and unpacks the filename. If, instead, the user links an existing (non-Magma-named) file on Metis, Magma uses that filename as the original_filename.

2. The user sends the original_filename as part of the link request to Magma. Magma copies the record via Metis and records the specified original_filename and the metis_path in the database. If there is no original_filename specified, Magma uses the metis_path to extract an original_filename.

coleshaw:
I was thinking to do option 2 of what you outlined above, @graft . I don't like overloading a file path with too much information (seems brittle), and since we're moving towards a JSON structure anyways to keep additional file data, it's easy to add in `original_filename` to the 2nd Magma request payload + the database.

coleshaw:
@graft , @corps is going to help cherry-pick some bug fixes for this branch and resolve the conflicts with master -- do you have any other feedback as we go through the process to get this merged in? That should get all the non-JSON-migration changes into Magma, so the breaking change will be smaller to deal with later.

lib/magma/attributes/file.rb
```diff
@@ -76,5 +79,10 @@ def filename(record_name, path)
       ext = '.dat' if ext.empty?
       "#{@model.model_name}-#{record_name}-#{@name}#{ext}"
     end
+
+    def temporary_filepath
+      Magma.instance.storage.upload_url(
+        @model.project_name, "tmp/#{SecureRandom.uuid}")
```
graft, line 85: Rather than using SecureRandom.uuid here you might prefer Magma.instance.sign.uid, which is usually how I generate random hex IDs in Etna apps; I prefer to centralize all of the cryptographic stuff in that place so all our potential mistakes are in one place.

lib/magma/server/update.rb
```diff
@@ -2,65 +2,179 @@
...
+            copy_revisions.push({
+              source: revision[attribute_name],
+              dest: "metis://#{@project_name}/magma/#{revision.to_loader[attribute_name][:filename]}"
+            })
+          end
+        end
       end
     end
+
+    if copy_revisions.length > 0 && Magma.instance.storage.instance_of?(Magma::Storage::Metis)
```
coleshaw, line 123: This storage check is to make the merge into master easier, until we make a full transition to Metis storage. Otherwise Magma would break the instance we merged this in, when using AWS storage.

Can remove the check at a later point, after the full transition.

coleshaw:
Superseded by #180

PR #161, opened 2020-06-03
[WIP] Issue143 file attribute json

coleshaw:
@graft , can I get your initial thoughts on if I did this correctly? I have a feeling this will break things on Timur too, since now Magma expects a JSON structure for file attributes, not just a string...

Also, can you look at `0.1.12` of Etna gem and merge / publish if it looks okay? This depends on that, so tests failing. mountetna/etna#50.


coleshaw:
Closing in favor of #180

PR #162, opened 2020-06-04
Load models from the newly added models table

### LINKED ISSUE #158: Create models table
Following #123  we added an attributes table, which current is able to completely load all of the attributes for a model; however, we do not yet load models themselves this way. If a model were a mere collection of attributes, we could certainly instantiate it using only the contents of the attributes table.

As it happens, there is already a single element that is defined on the model in addition to the attributes, a "dictionary", which defines a mapping between columns of this model and the columns of another model to be used as a dictionary for this one. This relationship is not at all like an attribute (a property of a record), and there is a complex datatype associated with it (the dictionary model name and the attribute map).

This by itself is justification for creating a separate "models" table in addition to the "attributes" table we already have. We might anticipate that in the future other properties of models might appear here (e.g., models might have a `model_type`, whatever that is). For the moment we can keep it simple:
```
  project_name: String
  model_name: String
  dictionary: json
```

As for whether we should normalize `model_name` out of the attributes table, it seems an unnecessary burden until the moment we might wish to start renaming models (a nightmare in its own right).

alimi:
There are a few changes we had to make to support loading models from the models table.

### Change `Magma::Model` inherited callback

The inherited callback in `Magma::Model` will only set the schema for models defined in Ruby. Models loaded from the database are created with anonymous classes, and anonymous classes don't have names. Therefore, we can't call methods like `#project_name` on them. Instead, we'll set the schema explicitly after the anonymous classes are initialized.

https://github.com/mountetna/magma/commit/4c5945bca0537595253e2045c9cb1481a700a87c

### Initialize `Magma::Dictionary` objects with the dictionary model name instead of the dictionary model class

This came up when deleting the `Labors::Codex` Ruby definition. Once the file was removed, `Labors::Aspect` was loaded from the database before `Labors::Codex`. As a result, `Labors::Aspect` tries to reference a non-existing `Labors::Codex` constant and blows up.

To prevent dictionary definitions from blowing up, update `Magma::Dictionary` so it's initialized with the dictionary model name instead of the dictionary model class. The dictionary model class *should* be loaded by the time we need it.

https://github.com/mountetna/magma/commit/d19b98e30787fadb0b6885961fd09865c0f94ad6

### Delete `Magma::Model#restricted`

`Labors::Victim` used the `Magma::Model#restricted` method to define a boolean attribute named `:restricted`. Instead of using `Magma::Model#restricted`, we can add the `:restricted` boolean attribute to `Labors::Victim` fixture.

https://github.com/mountetna/magma/commit/9c30e80f2a3ddf5537073ebeec36c1a08ef54fde

---

This PR also truncates the template tables (`attributes` and `models`) before test runs to ensure the tables are in a good state. https://github.com/mountetna/magma/commit/cb65b2615c0e1ae304eb635d288a4ed4b231731f

Fixes #158


lib/magma/project.rb
```diff
@@ -62,9 +62,27 @@ def load_project
...
+          model_class = Class.new(Magma::Model) do
+            set_schema(
+              model[:project_name].to_sym,
+              model[:model_name].pluralize.to_sym
+            )
+
+            dictionary(model[:dictionary].symbolize_keys) if model[:dictionary]
+          end
+
+          project_container.const_set(model[:model_name].classify, model_class)
```
alimi, line 82: By adding the model constant onto the `project_container`, the anonymous classes become more concrete and respond to methods like `name`.

```ruby
 => 82:           binding.irb
    83:           project_container.const_set(model[:model_name].classify, model_class)
    84:           binding.irb
    85:         end
    86:     end
    87: 

irb(#<Magma::Project:0x0000558bf3ff4fb8>):001:0> model_class
=> #<Class:0x0000558bf402e858>
irb(#<Magma::Project:0x0000558bf3ff4fb8>):002:0> model_class.name
=> nil
```

```ruby
    83:           project_container.const_set(model[:model_name].classify, model_class)
 => 84:           binding.irb
    85:         end
    86:     end
    87: 
    88:     def load_model_attributes
    89:       model_attributes = Magma.instance.db[:attributes].

irb(#<Magma::Project:0x0000558bf3ff4fb8>):001:0> model_class
=> Labors::Aspect
irb(#<Magma::Project:0x0000558bf3ff4fb8>):002:0> model_class.name
=> "Labors::Aspect"
```

alimi:
I also wonder if we could make dictionary an attribute and add it to the attributes table. I don't think it'd be too hard to do, and future changes like changing a model's name would be easier because we'd have one less table to worry about. What do you think @graft?

graft:
> I also wonder if we could make dictionary an attribute and add it to the attributes table. I don't think it'd be too hard to do, and future changes like changing a model's name would be easier because we'd have one less table to worry about. What do you think @graft?

I also thought about this and almost wrote that issue instead - the idea of making the model merely a collection of attributes and having a single table is clearly better. However, it doesn't work. The dictionary is a property of the model; it tells you how to map attributes between two models. An attribute is a property of a record; if we had a "dictionary" attribute, we would have to have a dictionary value for each record in the dataset.

There might be ways to salvage this, like attaching all of your dictionary attributes to the project model (which has a single record), but this seems pretty kludgy. We ought to admit that the dictionary exists at the model level. Also, as I said in #158 we will surely encounter other properties of models in the future (for example, we might want to globally restrict an entire model to privileged users), and we should make way for them by just setting up the models table now.

alimi:
👍🏾 makes sense.

alimi:
I wanted to clarify what's going on with the anonymous classes in [`Magma::Project#load_models`](https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/lib/magma/project.rb#L69-L83).

After we create an anonymous class for a model from the database

```ruby
          model_class = Class.new(Magma::Model) do
            # …
          end
```

we define a constant for it under [`Magma::Model#project_container`](https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/lib/magma/project.rb#L51-L53).

```ruby
          project_container.const_set(model[:model_name].classify, model_class)
```

This gives us classes like `Ipi::Patient` and `Labors::Monster` that we can use throughout the codebase, in the console, etc. The classes will respond to all the methods that we'd normally have available. We can see that in a test [that adds a `Labors::Hero` model to the models and attributes tables](https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/spec/magma_project_spec.rb#L95-L117).

```ruby
From: /home/developer/magma/spec/magma_project_spec.rb @ line 117 :

    112:
    113:       project = Magma::Project.new("./labors")
    114:
    115:       expect(project.models[:hero]).to eq(Labors::Hero)
    116:       expect(Labors::Hero.attributes[:name].description).to eq("The hero's name")
 => 117:       binding.irb
    118:     end
    119:   end
    120:
    121:   describe '.ordered_models' do
    122:     let(:project) { Magma.instance.magma_projects[:labors] }

irb:001:0> Labors::Hero.project_name
=> :labors
irb:002:0> Labors::Hero.model_name
=> :hero
```

We only run into a problem with the anonymous classes when they first get created because anonymous classes return `nil` for `#name`. Since the anonymous classes inherit from `Magma::Model`, the [`inherited` callback](https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/lib/magma/model.rb#L182-L197) will be called. The callback assumes all subclasses will have a value for `#name`, but the anonymous classes don't and the callback errors as a result.

We can get around the problems with the `Magma::Model` `inherited` callback by skipping the parts that depend on the anonymous classes having values for `#name`.

https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/lib/magma/model.rb#L182-L191

Then after the callback is called, we can manually call whatever depends on `#name`.

https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/lib/magma/project.rb#L73-L77

_We're working around `#name` being `nil` here, but [after the constant is set](https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/lib/magma/project.rb#L82) we could call  `set_schema` with `model_class#name`._

If `Magma::Model` didn't define an `inherited` callback, we'd probably never have run into problems with the anonymous classes not having names because the window where that happens is *really* short. The constant [gets set right after the anonymous class is created](https://github.com/mountetna/magma/blob/cb65b2615c0e1ae304eb635d288a4ed4b231731f/lib/magma/project.rb#L82).

Hopefully this explanation is helpful. I'm also happy to pair on this - it might be easier to see what happens in practice than reading this long comment 😛

graft:
:shipit:  I don't think there are any actual issues with the current behavior of the model-loading, since we have specs covering both `Magma::Project#load_models` and `Magma#validate_models`; these don't really need to be tested together.

We can probably improve how Magma warns on errors, and maybe have more permissive loading, but none of those issues are easily addressable here.

PR #163, opened 2020-06-08
Add validations for update model

notmarkmiranda:


spec/server/update_model_spec.rb
```diff
@@ -43,4 +46,42 @@ def app
...
+        display_name: "NAME",
+        hidden: 'something'
+      }]
+    })
+
+    response_json = JSON.parse(last_response.body)
+    expect(last_response.status).to eq(422)
+    expect(response_json['errors'][0]['message']).to eq("Update attribute failed")
+  end
 end
```
coleshaw, line 87: Would it be worth it to add specs that the other `EDITABLE_OPTIONS` are actually editable, or is that too much test fatigue? Or at least a spec around the validation piece that iterates through those?

graft: I agree about specing the validation piece, that is trouble.

Issue #131, opened 2020-04-29
Add /update_model endpoint


graft:
# Introduction

The current model development cycle in Magma goes like this:

First we plan:
1. Add new models or alter existing models in the project repo.
2. Load the new models into Magma; use the "plan" command to compute a migration, based on the difference between the current database schema and the current model defintion.
  
Then we modify the plan, i.e., add or remove atomic operations to the database
3. Check the planned migration for errors (notably, the plan cannot distinguish between "rename column" and "drop column" + "add column"). Rewrite to fix errors and allow other data loading tasks.
4. Run the migration.rb file to update the database.

Subsequently there is cleanup and deploy:
5. Validate the model, perhaps by test-loading some data and building views in Timur
6. Push code (models + migration changes) to production

There are two broad strategies we might use to exposing model-editing functionality:

1) Describe the new model, compute the differences required and make the appropriate changes
This "plan" methodology allows us to quickly dump in a large template with perhaps hundreds of attributes. However, the plan is not trustworthy. While using it allows us to sketch in broad strokes, the actual migrations it generates must be validated and corrected before being run. This means the process above is dangerous: that is, it is not *guaranteed* to never accidentally destroy data. This requirement for validation of migrations also slows down the process of deploying new models.

2) Allow atomic operations to be executed by the user
This allows us to sculpt in details, adding or removing attributes to reshape our existing models. But it would grow tedious to sketch out new projects this way, adding in attributes to a model one-at-a-time.
    
Here are some scenarios we might imagine using these operations:
- We create a new project. We wish to quickly inflate a data graph that we can put data in; we already know the data types and structure. We assemble a basic set of models using templates.
- We have an existing project, which is now processing samples using a new data type. We add in models for this data type to the existing project data graph based on some templates.
- Our existing project requires some new attributes to support an idiosyncratic data requirement. We add the new attributes to an existing model.
- One of our attributes is improperly harboring PHI. We wish to remove it from the model and destroy the data.

The basic operations we might require are:

1) add a model, e.g. add_model { model_name: 'sample', parent_model: 'patient', and parent_link_type: 'collection' }
2) remove a model, e.g. remove_model { model_name: 'sample' }
3) add an attribute to a model, e.g. add_attribute { model_name: 'sample', attribute_name: 'generation_time', ...attribute_params }
4) remove an attribute from a model, e.g. remove_attribute { model_name: 'sample', attribute_name: 'generation_time' }
[...]

PR #168, opened 2020-06-10
Add /update_model API endpoint

alimi:
This PR adds a /update_model API endpoint per #131. So far it supports updating an attribute's options.

The endpoint is backed by the newly added `UpdateModelController`. It will iterate over the array of actions it receives, and defer to action classes to perform those actions. For example, an `update_attribute` action will be performed by `Magma::UpdateAttribute`.

In addition to performing actions, each action class is responsible for validating its action params. This division of responsibilities will help keep `UpdateModelController` pretty stable as we add new actions. It will also make it easier for us to parallelize the work going forward.

We've also restricted the endpoint to superusers based on conversations we've had in the past. We can change the permissions if something else makes sense.

### DIFF
--- a/lib/magma/actions/action_error.rb
+++ b/lib/magma/actions/action_error.rb
@@ -1,0 +1,15 @@
+class Magma
+  class ActionError
+    attr_reader :message, :source, :reason
+
+    def initialize(message:, source:, reason: nil)
+      @message = message  
+      @source = source
+      @reason = reason
+    end
+
+    def to_h
+      {message: message, source: source, reason: reason}
+    end
+  end
+end
--- a/lib/magma/actions/update_attribute.rb
+++ b/lib/magma/actions/update_attribute.rb
@@ -31,6 +31,8 @@ class Magma
     def validate
       if attribute
         @action_params.except(:action_name, :model_name, :attribute_name).keys.all? do |option|
+          next if check_restricted_attributes(option)
+
           unless attribute.respond_to?(option)
             @errors << Magma::ActionError.new(message: "Attribute does not implement #{option}", source: @action_params.slice(:action_name, :model_name, :attribute_name))
           end
@@ -43,6 +45,16 @@ class Magma
 
     def errors
       @errors.map(&:to_h)
+    end
+
+    private
+
+    def check_restricted_attributes(option)
+      if option == :name || option == :new_attribute_name
+        @errors << Magma::ActionError.new(message: "#{option.to_s} cannot be changed", source: @action_params.slice(:action_name, :model_name, :attribute_name))
+        true
+      end
+      false
     end
   end
 end
--- a/lib/magma/actions/update_attribute.rb
+++ b/lib/magma/actions/update_attribute.rb
@@ -1,7 +1,7 @@
 require 'magma/actions/action_error'
 
 class Magma
-  class UpdateAttribute
+  class UpdateAttributeAction
     def initialize(project_name, action_params = {})
       @project_name = project_name
       @action_params = action_params
--- a/lib/magma/actions/update_attribute.rb
+++ b/lib/magma/actions/update_attribute.rb
@@ -1,14 +1,20 @@
+require 'magma/actions/action_error'
+
 class Magma
   class UpdateAttribute
     def initialize(project_name, action_params = {})
       @project_name = project_name
       @action_params = action_params
+      @errors = []
     end
 
     def perform
       @action_params.slice(*Magma::Attribute::EDITABLE_OPTIONS).each do |option, value|
         attribute.update_option(option, value)
+      rescue => e
+        @errors << Magma::ActionError.new(message: 'Update attribute failed', source: @action_params.slice(:attribute_name, :model_name), reason: e)
       end
+      return @errors.empty?
     end
 
     def model
@@ -23,10 +29,20 @@ class Magma
     end
 
     def validate
-      return false unless attribute
-      @action_params.except(:action_name, :model_name, :attribute_name).keys.all? do |option|
-        attribute.respond_to?(option)
+      if attribute
+        @action_params.except(:action_name, :model_name, :attribute_name).keys.all? do |option|
+          unless attribute.respond_to?(option)
+            @errors << Magma::ActionError.new(message: "Attribute does not implement #{option}", source: @action_params.slice(:action_name, :model_name, :attribute_name))
+          end
+        end
+      else
+        @errors << Magma::ActionError.new(message: 'Attribute does not exist', source: @action_params.slice(:attribute_name, :model_name))
       end
+      return @errors.empty?
+    end
+
+    def errors
+      @errors.map(&:to_h)
     end
   end
 end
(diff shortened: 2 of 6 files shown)


lib/magma/actions/update_attribute.rb
```diff
@@ -0,0 +1,48 @@
+require 'magma/actions/action_error'
+
+class Magma
+  class UpdateAttribute
+    def initialize(project_name, action_params = {})
```
graft, line 5: Could this be like `Magma::UpdateAttributeAction` or `Magma::Action::UpdateAttribute`?

spec/server/update_model_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        attribute_name: proper_name,
+        description: 123,
+        display_name: "NAME",
+        hidden: 'something'
+      }]
+    })
+
+    response_json = JSON.parse(last_response.body)
+    expect(last_response.status).to eq(422)
+    expect(response_json['errors'][0]['message']).to eq("Update attribute failed")
```
graft, line 85: A clearer message here might help the user correct the request.

graft:
Pretty much exactly what I imagined. It seems there is as of yet no base `Magma::Action` type class (is that the right name?), I guess it will show up later.

As a sanity check I might verify that this thing refuses to change the attribute name and type.

Otherwise... can this just be merged? o_O

spec/spec_helper.rb
```diff
@@ -286,6 +286,9 @@ def json_document model, record_name
 end
 
 AUTH_USERS = {
+  superuser: {
+    email: 'zeus@twelve-labors.org', first: 'Zeus', perm: 'a:administration'
```
coleshaw, line 290: Should the permissions on this new user be for the `labors` project instead of `administration`?

spec/server/update_model_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        description: "The monster's name",
+        display_name: "NAME"
+      }]
+    })
+
+    expect(last_response.status).to eq(200)
+    expect(Labors::Monster.attributes[:name].description).to eq("The monster's name")
+    expect(Labors::Monster.attributes[:name].display_name).to eq("NAME")
+
+    response_json = JSON.parse(last_response.body)
```
coleshaw, line 44: nit: there is a `json_body()` method in `spec_helpers` that does this parsing of `last_response` for you, so that could save some writing in these specs.

lib/magma/actions/action_error.rb
```diff
@@ -0,0 +1,15 @@
+class Magma
+  class ActionError
```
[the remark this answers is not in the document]
alimi: A Struct could make a lot of sense here, but I'd recommend against subclassing it. It goes against the intended use of Structs

> ``` 
> Customer = Struct.new(:name, :address) do
>   def greeting
>     "Hello #{name}!"
>   end 
> end 
> Customer.new("Dave", "123 Main").greeting  #=> "Hello Dave!"
> ``` 
> 
> This is the recommended way to customize a struct. Subclassing an anonymous struct creates an extra anonymous class that will never be used.

https://ruby-doc.org/core-2.7.1/Struct.html

and it could lead to strange behavior down the road.

_I'm an expert on this topic because my wife [gave a talk](http://latazzajones.github.io/tashas_presentations/ruby_structs.html#/_) on it three years ago 😆_

spec/spec_helper.rb
alimi: `administration` makes the user a superuser.

```ruby
module Etna
  class User
    # …
    def is_superuser? project=nil
      has_roles(:administration, :admin)
    end
    # …
  end
end
```

https://github.com/mountetna/etna/blob/089c179a66f3b451a82cea8ddec133f738f6948e/lib/etna/user.rb#L55-L57

If we changed it to `labors`, the user would only have admin access on the Labors project.

coleshaw: Ah interesting. Thanks for clarifying!

spec/server/update_model_spec.rb
notmarkmiranda: @graft, we capture the error message that comes back from Sequel on the insert failure: 
```
{"message"=>"Update attribute failed",
 "source"=>{"attribute_name"=>"name", "model_name"=>"monster"},
 "reason"=>
  "PG::InvalidTextRepresentation: ERROR:  invalid input syntax for type boolean: \"something\"\nLINE 1: ....024405-0600', '2020-06-11 09:04:33.024406-0600', 'something...\n                                                             ^\n"}
```

I'll update the spec to reflect that.

lib/magma/actions/update_attribute.rb
alimi: We renamed it to `Magma::UpdateAttributeAction` in https://github.com/mountetna/magma/pull/168/commits/05840ff7378d0818f27eb2ca9ce31623748549ee. It follows the same pattern we use for attributes - they live in `lib/magma/attributes` and are named like `Magma::BooleanAttribute`.

lib/magma/actions/update_attribute.rb
```diff
@@ -0,0 +1,48 @@
...
+      @project_name = project_name
+      @action_params = action_params
+      @errors = []
+    end
+
+    def perform
+      @action_params.slice(*Magma::Attribute::EDITABLE_OPTIONS).each do |option, value|
+        attribute.update_option(option, value)
+      rescue => e
```
graft, line 14: This broad rescue will probably get us into trouble; here we are catching everything and sending it back as a 422, "Bad Request". This may be correct, but the exception may instead be due to an error on our part, which should return as a 500. If we intend to use this rescue to catch 422 errors, we should rescue precisely only those errors that could be the result of user error and not just any exception. Although in general, I think my preference is usually to check for user errors by hand rather than relying on an exception being generated.

You are also packaging the message from the exception and returning it to the user. The raw error is useful for debugging and could go into the logs, but the user is not debugging the code, and does not know or need to know what the ruby class of the error was, etc. (in addition to whatever else the error might improperly reveal, who knows). For this reason I'd rather give the user a formatted response based on the error type, tailored to help them turn their "bad request" into a "good request", but not the raised exception itself.

spec/server/update_model_spec.rb
graft: Argh i wrote a whole thing about this earlier but it didn't post? Anywho see above.

notmarkmiranda: About rescuing all errors and the necessity of differentiating between a 422 & 500? If so, yes: https://github.com/mountetna/magma/pull/168#discussion_r438913270

lib/magma/actions/update_attribute.rb
notmarkmiranda: I think this should be addressed in a separate PR. I think the best way to accomplish this is with model-level validations on the attribute. We could

alimi:
If everyone is agreeable, I think we should merge in this PR. It sets up a decent foundation for us to build off of, and we can avoid having a long running PR.

graft:
Sure, let's merge to have a solid footing.

PR #174, opened 2020-06-15
Test fixes

alimi:
https://github.com/mountetna/magma/commit/f56a5c0406b7ba7ebaf178307d6403abcd11bd07 resets attributes that get changed in tests. `Magma::Attribute#update_option` changes options in memory that persist between test runs. As a result, attribute changes in `UpdateModelController` spec can cause other tests to fail. For example, prior to this commit `rspec spec/server/update_model_spec.rb spec/magma_commands_spec.rb --seed 18642` would fail. Before we tackle the bigger challenges around global state, we can manually reset attributes that get changed.

https://github.com/mountetna/magma/commit/5fcbebab11292527cc723ef7424b9164898f3857 adds a new fixtures project called Movies that can be used to verify the behavior of projects with Ruby backed model definitions. This helps give us more confidence that production projects like Ipi that are still Ruby backed work.

### DIFF
--- a/spec/fixtures/movies/models/villain.rb
+++ b/spec/fixtures/movies/models/villain.rb
@@ -1,0 +1,5 @@
+module Movies
+  class Villain < Magma::Model
+    string :name, description: "The villain's name"
+  end
+end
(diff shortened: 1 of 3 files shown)


graft:
:+1: ship it quickly, before it becomes obsolete.

PR #177, opened 2020-06-22
Make Magma::Attribute inherit from Sequel::Model

alimi:
By making `Magma::Attribute` inherit from `Sequel::Model`, we can leverage Sequel's API to manage data we get from the attributes table. This will help us avoid issues like https://github.com/mountetna/magma/pull/173#issuecomment-644144794 that come up from us trying to manage everything ourselves.

To get this to work, we have to make a number of changes. A few that stand out

  * Configure Sequel after connecting to the database so Sequel is available in both `Magma::Attribute` and `Magma::Model`
  * Use Sequel's [single table inheritance](http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/SingleTableInheritance.html) plugin to support `Magma::Attribute` subclasses (`Mamga::StringAttribute`, `Magma::IntegerAttribute`, etc.)
  * Rename `Magma::Attribute#model` to `#magma_model` because `#model` is used by `Sequel::Model`
  * Add `Magma::Attribute#after_magma_model_set` hook method because `Magma::Attribute`s fetched from the database don't get `magma_model`s when they're retrieved from the database.
  * Rename `Magma::Attribute#update` to `#update_record` because `#update` is used by `Sequel::Model`
  * Set `project_name` and `model_name` on `Magma::Attribute`s defined in Ruby. This makes it so we can call `#update` later to make changes.
    * To facilitate this, we can't set timestamp attributes in the `Magma::Model` inherited hook. https://github.com/mountetna/magma/commit/d1ac05a89092192f64085269b36114e3d5a9388c

This PR also makes some changes to loading the Labors project in `spec/spec_helper.rb` to make it clearer and more consistent.

Everything boots up with these changes locally, and I can click through the parts of Timur I know to make sure everything is working as it was before. I'd appreciate some more eyes to make sure I haven't broken something I don't know about 🙈

### DIFF
--- a/lib/magma.rb
+++ b/lib/magma.rb
@@ -40,17 +40,11 @@ class Magma
     @db.extension :connection_validator
     @db.extension :pg_json
     @db.pool.connection_validation_timeout = -1
-
-    Sequel::Model.plugin :timestamps, update_on_create: true
-    Sequel::Model.require_valid_table = false
-    Sequel.extension :inflector
-
-    require_relative 'magma/attribute'
-    require_relative 'magma/model'
   end
 
   def load_models(validate = true)
     setup_db
+    setup_sequel
 
     @storage = Magma::Storage.setup
 
@@ -60,6 +54,14 @@ class Magma
     end
 
     validate_models if validate
+  end
+
+  def setup_sequel
+    Sequel::Model.plugin :timestamps, update_on_create: true
+    Sequel::Model.require_valid_table = false
+    Sequel.extension :inflector
+    require_relative 'magma/attribute'
+    require_relative 'magma/model'
   end
 
   class Magma::ValidationError < StandardError
--- a/lib/magma.rb
+++ b/lib/magma.rb
@@ -6,9 +6,7 @@ require_relative 'magma/validation'
 require_relative 'magma/validation_object'
 require_relative 'magma/loader'
 require_relative 'magma/migration'
-require_relative 'magma/attribute'
 require_relative 'magma/dictionary'
-require_relative 'magma/model'
 require_relative 'magma/revision'
 require_relative 'magma/commands'
 require_relative 'magma/payload'
@@ -42,6 +40,13 @@ class Magma
     @db.extension :connection_validator
     @db.extension :pg_json
     @db.pool.connection_validation_timeout = -1
+
+    Sequel::Model.plugin :timestamps, update_on_create: true
+    Sequel::Model.require_valid_table = false
+    Sequel.extension :inflector
+
+    require_relative 'magma/attribute'
+    require_relative 'magma/model'
   end
 
   def load_models(validate = true)
--- a/lib/magma/actions/update_attribute.rb
+++ b/lib/magma/actions/update_attribute.rb
@@ -10,7 +10,16 @@ class Magma
 
     def perform
       attribute.update(@action_params.slice(*Magma::Attribute::EDITABLE_OPTIONS))
-      @errors.empty?
+    rescue Sequel::ValidationFailed => e
+      @errors << Magma::ActionError.new(
+        message: 'Update attribute failed',
+        source: @action_params.slice(:attribute_name, :model_name),
+        reason: e
+      )
+
+      attribute.initial_values.keys.each { |name| attribute.reset_column(name) }
+    ensure
+      return @errors.empty?
     end
 
     def model
--- a/lib/magma/actions/update_attribute.rb
+++ b/lib/magma/actions/update_attribute.rb
@@ -9,12 +9,8 @@ class Magma
     end
 
     def perform
-      @action_params.slice(*Magma::Attribute::EDITABLE_OPTIONS).each do |option, value|
-        attribute.update_option(option, value)
-      rescue => e
-        @errors << Magma::ActionError.new(message: 'Update attribute failed', source: @action_params.slice(:attribute_name, :model_name), reason: e)
-      end
-      return @errors.empty?
+      attribute.update(@action_params.slice(*Magma::Attribute::EDITABLE_OPTIONS))
+      @errors.empty?
     end
 
     def model
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -43,10 +43,6 @@ class Magma
 :format_hint, :loader, :link_model_name, :restricted]
       end
 
-      def set_attribute(name, model, options, attribute_class)
-        attribute_class.new(name, model, options)
-      end
-
       def attribute_type
         @attribute_type ||= name.match("Magma::(.*)Attribute")[1].underscore
       end
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -34,7 +34,6 @@ class Magma
       :validation
     ]
 
-    attr_accessor :magma_model
     attr_reader :loader
 
     class << self
@@ -49,11 +48,20 @@ class Magma
     end
 
     def initialize(opts = {})
-      @magma_model = opts.delete(:magma_model)
-      @loader = opts.delete(:loader)
       # Some Ipi models set the group option but it doesn't seem to be used anywhere
       opts.delete(:group)
-      super
+
+      super(opts.except(:magma_model, :loader))
+      self.magma_model = opts[:magma_model]
+      @loader = opts[:loader]
+    end
+
+    def magma_model=(new_magma_model)
+      @magma_model = new_magma_model
+      after_magma_model_set
+    end
+
+    def after_magma_model_set
     end
 
     def name
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -1,5 +1,24 @@
 class Magma
-  class Attribute
+  class Attribute < Sequel::Model
+    plugin :single_table_inheritance, :type, model_map: {
+      "parent" => "Magma::ParentAttribute",
+      "string" => "Magma::StringAttribute",
+      "match" => "Magma::MatchAttribute",
+      "identifier" => "Magma::IdentifierAttribute",
+      "child" => "Magma::ChildAttribute",
+      "integer" => "Magma::IntegerAttribute",
+      "boolean" => "Magma::BooleanAttribute",
+      "date_time" => "Magma::DateTimeAttribute",
+      "table" => "Magma::TableAttribute",
+      "matrix" => "Magma::MatrixAttribute",
+      "collection" => "Magma::CollectionAttribute",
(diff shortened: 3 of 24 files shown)


lib/magma/attribute.rb
```diff
@@ -15,45 +34,56 @@ class Attribute
...
       end
     end
 
-    def initialize(name, model, opts)
-      @name = name
-      @model = model
-      set_options(opts)
+    def initialize(opts = {})
+      # Some Ipi models set the group option but it doesn't seem to be used anywhere
+      opts.delete(:group)
```
alimi, line 52: Like the comment says, some Ipi models set the `group` option but I didn't find any uses of it in magma. Do we need to support it?

lib/magma/attribute.rb
```diff
@@ -15,45 +34,56 @@ class Attribute
...
+
+      super(opts.except(:magma_model, :loader))
+      self.magma_model = opts[:magma_model]
+      @loader = opts[:loader]
```
alimi, line 56: We set `@loader` for backwards compatibility with Ruby backed attributes in Ipi.

lib/magma/attribute.rb
```diff
@@ -15,45 +34,56 @@ class Attribute
...
+    def magma_model=(new_magma_model)
+      @magma_model = new_magma_model
+      after_magma_model_set
+    end
+
+    def after_magma_model_set
+    end
+
+    def name
+      attribute_name.to_sym
```
alimi, line 68: There are a few places in the codebase that need `attribute_name` to be a symbol, so I left this here as a convenience method. I'd be open to removing it though.

lib/magma/attribute.rb
```diff
@@ -125,40 +154,8 @@ def entry
       end
     end
 
-    def update_option(opt, new_value)
```
alimi, line 128: Thanks to `Sequel::Model` we don't have to write/use things like `Magma::Attribute#update_option` 🎉. This also means we don't have to deal with the `INSERT CONFLICT` weirdness @graft was seeing.

spec/server/update_model_spec.rb
```diff
@@ -74,31 +74,6 @@ def app
       Labors::Monster.attributes[:name] = original_attribute
     end
 
-    it "does not update attribute options with the incorrect data type" do
```
alimi, line 77: This test isn't relevant anymore because it's not possible to set options with incorrect data types. `Sequel::Model` will cast values so they can be inserted in the database. This makes it easier to do things like insert a Ruby hash into a JSON column.

If we wanna do this kind of validation, we can look into `Sequel::Model`s type validation. https://sequel.jeremyevans.net/rdoc/files/doc/validations_rdoc.html#label-validates_type

lib/magma/attribute.rb
```diff
@@ -15,45 +34,56 @@ class Attribute
...
-    attr_reader :name, :loader, :validation, :format_hint, :unique, :index, :restricted, :link_model_name, :description
+    attr_reader :loader
 
     class << self
       def options
         [:description, :display_name, :hidden, :read_only, :unique, :index, :validation,
 :format_hint, :loader, :link_model_name, :restricted]
       end
 
-      def set_attribute(name, model, options, attribute_class)
```
alimi, line 26: `#set_attribute` doesn't seem to be used anywhere.

lib/magma.rb
```diff
@@ -42,6 +40,13 @@ def setup_db
     @db.extension :connection_validator
     @db.extension :pg_json
     @db.pool.connection_validation_timeout = -1
+
+    Sequel::Model.plugin :timestamps, update_on_create: true
+    Sequel::Model.require_valid_table = false
+    Sequel.extension :inflector
+
```
graft, line 47: Will this play well with ActiveSupport, which also tries to do pluralization? I've run into trouble here before due to confusion over who, exactly, is determining what is "plural" at any given time, and differing ideas of the correct pluralization.

lib/magma/actions/update_attribute.rb
```diff
@@ -9,12 +9,8 @@ def initialize(project_name, action_params = {})
...
 
     def perform
-      @action_params.slice(*Magma::Attribute::EDITABLE_OPTIONS).each do |option, value|
-        attribute.update_option(option, value)
-      rescue => e
-        @errors << Magma::ActionError.new(message: 'Update attribute failed', source: @action_params.slice(:attribute_name, :model_name), reason: e)
-      end
-      return @errors.empty?
+      attribute.update(@action_params.slice(*Magma::Attribute::EDITABLE_OPTIONS))
+      @errors.empty?
```
graft, line 13: It's not clear to me now how @errors gets populated

lib/magma/attribute.rb
graft: We're interested in using this more in the future to help organize the attribute space; but we haven't really developed the notion fully so it can be left aside for the moment.

lib/magma/model.rb
```diff
@@ -1,30 +1,27 @@
...
+  class Model < Sequel::Model
     class << self
       Magma::Attribute.descendants.
         reject { |attribute| attribute == Magma::ForeignKeyAttribute }.
         each do |attribute|
           define_method attribute.attribute_type do |attribute_name=nil, opts={}|
             @parent = attribute_name if attribute == Magma::ParentAttribute
-            attributes[attribute_name] = attribute.new(attribute_name, self, opts)
+            attributes[attribute_name] = attribute.new(opts.merge(
+              project_name: project_name,
```
graft, line 10: What is the behavior of 'new' and its relation to the (possibly non-existent) db entry here?

graft:
What a lot of work! This is a very valuable change that seems to clear up a lot of confusing divergences and logic-duplication in the code.

One thing I am curious about is: previously there were two separate classes that allowed the separation of the attribute from its database entity, which was useful in the situation where no database entity existed yet to describe that attribute (i.e., it came from a .rb model). It seems like this situation persists with `Magma::Attribute.new()`, which creates an attribute object but does not create the corresponding record?

lib/magma/actions/update_attribute.rb
alimi: `@errors` will get set when call `validate` on an action.

https://github.com/mountetna/magma/blob/ef767145c933f0186474a3efce27b7170693be8e/lib/magma/server/update_model.rb#L11

https://github.com/mountetna/magma/blob/ef767145c933f0186474a3efce27b7170693be8e/lib/magma/actions/update_attribute.rb#L31-L44

Before `#perform` could also populate `@errors` if an error happened during the attribute update. Now that we're using `Sequel::Model#update` to update attributes, exceptions are a lot less likely to happen. I tried to make it error but was unsuccessful.

We _probably_ don't need to return `@errors.empty?` from `#perform` anymore - it just needs to return a truthy value.

lib/magma/model.rb
alimi: From your comment below

> It seems like this situation persists with Magma::Attribute.new(), which creates an attribute object but does not create the corresponding record?

That's exactly what's happening. With `Magma::Attribute.new()`, we can load non-db backed attributes in memory. And then if we want to make changes, we can call `Magma::Attribute.update()` to insert that row into the attributes table.

lib/magma.rb
alimi: Hmm...lemme play around with this a little bit. I didn't run into any problems when I was building this out and testing the changes.

alimi:
Per our discussion in stand up yesterday, I added some `Sequel::Model` validation checks in https://github.com/mountetna/magma/pull/177/commits/9527753eb747c8d8c187c1cee440b0ebb18cb038. 

There's an [`auto_validations` plugin](https://sequel.jeremyevans.net/rdoc/files/doc/validations_rdoc.html#label-auto_validations) that adds a bunch of validations based on the db schema. It will catch stuff like not null violations and present the errors back in user friendly messages. It still won't catch a lot of the type errors because the validations are based on the type casted values. The validation JSON seemed the most likely to be abused by this, so I added a custom validation that will check if the validation_object can be built from the given JSON. If the validation_object can't be built, we'll add a validation error on the attribute.

https://github.com/mountetna/magma/blob/9527753eb747c8d8c187c1cee440b0ebb18cb038/lib/magma/attribute.rb#L30-L35

It's not perfect but it might be good enough for now.

I also brought in Sequel's [`dirty` plugin](https://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/Dirty.html), so we can roll back changes in memory when an update fails. If an update fails due to a validation error, no changes will be made to the database. However, the objects in memory will have the changed values. The `dirty` plugin keeps track of the changes but we have to roll them back manually.

https://github.com/mountetna/magma/blob/9527753eb747c8d8c187c1cee440b0ebb18cb038/lib/magma/actions/update_attribute.rb#L11-L23

alimi:
Hmm...so the `auto_validations` plugin doesn't work when we're trying to set up a new database. When we use the `Magma::CreateDb` command to create the database and project schema, we call `Magma.instance.setup_db` to set up a database connection. [`#setup_db` requires `magma/attribute`](https://github.com/mountetna/magma/blob/9527753eb747c8d8c187c1cee440b0ebb18cb038/lib/magma.rb#L48) which then loads the `auto_validations` plugin on `Magma::Attribute`.

https://github.com/mountetna/magma/blob/9527753eb747c8d8c187c1cee440b0ebb18cb038/lib/magma/attribute.rb#L23

`auto_validations` queries the database to get info about the attributes table so it can add validations. When the `Magma::CreateDb` command is setting up a new database, the attributes table won't exist yet and the command blows up. That's what happens in the test run here: https://github.com/mountetna/magma/pull/177/checks?check_run_id=808661893#step:7:11.

I tried moving things around, but when we're creating a new database the migrations haven't been run yet so there's no way the attributes table could exist!

It'd be nice to use the `auto_validations` plugin but it might be more trouble than it's worth. I'm curious if anyone has any ideas; otherwise, I'm gonna remove it.

graft:
Could you run the global migrations after creating the db and before
running setup_db?

alimi:
Ok, I got it figured out in https://github.com/mountetna/magma/pull/177/commits/8166d4372e6dc492458805af2d03657da7e70597 ☺️. Instead of doing all the `Sequel::Model` setup in `Magma.setup_db`, we can defer it until `Magma.load_models`. This lets us get a db connection from `Magma.setup_db` without having to worry about the order in which things are loaded/set up.

I also made a change for the dirty plugin in https://github.com/mountetna/magma/pull/177/commits/94a341e12eb1640c92fce66e0eba5a525e8a7646.

PR #179, opened 2020-06-23
Load project models in the load_project command

### LINKED ISSUE #150: Add load_project command
In #148 we gained the ability to load attributes from the attributes table. How is the attributes table populated? In the specs, the example `labors` project is being loaded from a YAML fixture.

However, we already have a document format that attempts to describe the attributes for a project model completely:  `Magma::Model#json_template` and the associated `Magma::Attribute#json_template`; These are usually returned by the Magma /retrieve endpoint, and are a standard tool for describing magma models (used, for example, by the `ModelMap` or the `AttributeViewer` in Timur). We ought to be able to load attributes directly from these templates into a Magma instance's attributes (and models?) table.

This would create a nice loop, where we can dump templates from one project or instance and load them into another. These project templates can be paired with associated migrations to allow easier deploy of new models. (Eventually we ought to be able to load + inflate the data graph, i.e., auto-migrate, this way.)

The command would simply take a template.json as might be output by the /retrieve endpoint as an argument, and would make entries into the `attributes` table using each model/attribute's JSON definition.

alimi:
There's a little weirdness around loading model dictionaries from the template file because the model dictionary JSON is represented differently in the models table and the models JSON template.

In the models table, it looks like

https://github.com/mountetna/magma/blob/ef767145c933f0186474a3efce27b7170693be8e/spec/fixtures/labors_model_attributes.yml#L2

and in the template file it looks like

https://github.com/mountetna/magma/blob/0f5b7daf56417fc35fd0b60dbdd670b524d0c831/spec/fixtures/template.json#L188-L198

I added `dictionary_model` to the template so we can properly insert model dictionaries in `load_project`. So we can work around the dictionary JSON formats being different, but it might be worth consolidating it so it's consistent.

This PR also does a little bit of clean up on the command. It also updates the test so it runs against a new project that isn't already loaded by the test suite to make sure we're testing the command's behavior. 

Fixes #150

### DIFF
--- a/lib/magma/commands.rb
+++ b/lib/magma/commands.rb
@@ -9,47 +9,51 @@ class Magma
       file = File.open(file_name)
       file_data = JSON.parse(file.read, symbolize_names: true)
 
-      db = Magma.instance.db
-      models = file_data[:models]
+      file_data[:models].each do |model_name, model_json|
+        model_name = model_name.to_s
+        template = model_json[:template]
 
-      models.keys.each do |model|
-        model_name = models[model][:template][:name]
+        load_model(project_name, model_name, template)
 
-        dictionary_json = models[model][:template][:dictionary]
-
-        dictionary_data = if dictionary_json
-          dictionary_json[:attributes].merge(dictionary_model: dictionary_json[:dictionary_model])
-        else
-          nil
-        end
-
-        db[:models].insert(
-          project_name: project_name,
-          model_name: model_name,
-          dictionary: Sequel.pg_json_wrap(dictionary_data),
-        )
-
-        attributes = models[model][:template][:attributes]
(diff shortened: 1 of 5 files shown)


alimi:
I also forgot to mention - the `load_project` command assumes a project hasn't already been loaded into the models/attributes tables. It would blow up if there's already data in those tables for the project. I'm not sure if we want to make the command more resilient, but if we do it might be best for another issue down the road.

graft:
It's probably fine if it blows up, so long as it doesn't blow up halfway through an insertion. However, since there are uniqueness guarantees on both models and attributes tables it seems unlikely to cause too much trouble. In the end it seems like the behavior of the command should be idempotent, but the useful lifetime of this command is hopefully only a month or two anyway, so :foot: :football:

PR #180, opened 2020-06-26
Issue143 file attribute json

coleshaw:
This is a large PR that merges in work required for Magma-Metis integration (mountetna/planning#30).

There are several large changes included:

1. Instead of making record updates via individual Sequel transactions, we use the `Magma::Loader` class, which groups revisions together and does bulk inserts / updates. So this involves removing the `update` methods and replacing that with `revision_to_loader` and `revision_to_payload`. #40 

2. Different serialization methods -- `query_to_tsv` and `query_to_payload` instead of `json_for` or `tsv_for`.

3. Changing `FileAttribute` and `ImageAttribute` to JSON instead of just String. #143 

4. Adding new functionality on the `update` route to make a call to Metis for copying files. #51


coleshaw:
@graft @alimi , there is the Magma-Metis integration work on top of the latest `master` code, that includes PR# 177. Tests pass and the apps seem to work for me locally. Could you please give this a glance to see if I changed any of the model / attribute migration stuff incorrectly? Thanks!

spec/magma_attribute_spec.rb
```diff
@@ -77,6 +70,102 @@
...
+    it "returns entry for editable JSON backed options" do
+      attribute = Magma::MatchAttribute.create(
+        project_name: :project,
+        attribute_name: "name", 
+        model_name: :model,
+        validation: { type: "Array", value: [1, 2, 3] })
+
+      # Call attribute#validation_object to verify the cached validation_object
+      # gets reset
+      attribute.validation_object
```
alimi, line 95: We probably don't need this anymore since `#validation_object` isn't cached anymore.

spec/magma_child_attribute_spec.rb
```diff
@@ -12,8 +12,8 @@
         .and_return true
     end
 
-    it 'calls update or create on the link_model' do
-      child_attribute.update_record(record, link)
+    xit 'calls update or create on the link_model' do
```
alimi, line 15: Is this test supposed to be skipped?

spec/magma_file_attribute_spec.rb
```diff
@@ -0,0 +1,150 @@
+require_relative '../lib/magma'
+
+describe Magma::FileAttribute do
+  it 'stores a JSON database type' do
+    file_attribute = Magma.instance.magma_projects[:labors].models[:monster].attributes[:stats]
```
alimi, line 5: You could replace this with

```ruby
file_attribute = Labors::Monster.attributes[:stats]
```

Not a huge difference but it saves a few keystrokes 😄

spec/retrieve_spec.rb
```diff
@@ -1,3 +1,5 @@
+require 'pry'
```
alimi, line 1: Does pry need to be required here?

spec/storage_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'pry'
```
alimi, line 1: Does pry need to be required here too?

spec/magma_child_attribute_spec.rb
coleshaw: I should just remove this test -- the method doesn't exist any more on the attributes, and link updates wouldn't be handled by the `update_or_create` method anyways...

lib/magma/attribute.rb
```diff
@@ -116,57 +137,36 @@ def json_template
...
+    def query_to_tsv(value)
+      query_to_payload(value)
     end
 
-    def shown?
-      !hidden
+    def revision_to_loader(record_name, new_value)
+      [
+        name,
+        database_type == DateTime ?
```
corps, line 151: To reduce the ternary operator soup you can use
```
case database_type
when DateTime
   ...
when Integer
  ...
when Float
   ..

end
```

corps: Also, being able to convert a new value based on the attribute's database_type seems like a generally useful function to break out as separate function?

lib/magma/attributes/child.rb
```diff
@@ -1,14 +1,14 @@
...
+    def query_to_payload(value)
+      value
     end
 
-    def update_record record, link
-      link_model.update_or_create(link_model.identity => link) do |obj|
-        obj[ self_id ] = record.id
-      end
+    def revision_to_record(record, value)
+    end
```
corps, line 9: Are these supposed to be abstract methods (subclasses should impelement)?
If so, might be useful to indicate this with a comment or a `raise NotImplementedError`

lib/magma/attribute.rb
graft: I think that is the (only) purpose of this function, it bizarrely returns a [ name,  ... ] wrapped around the value => entry conversion because of the poor interface design/coupling with the corresponding usage of this method by the update endpoint to pass data between Revision and Loader, attribute-specifically. I wasn't really able to solve this problem very well, thus this ugly mess.

lib/magma/attributes/child.rb
graft: There is an implicit nil which I think gets compacted away.

lib/magma/censor.rb
```diff
@@ -0,0 +1,41 @@
+class Magma
+  class Censor
+    attr_reader :model, :record_name
+    def initialize(user, project_name)
+      @user = user
+      @project_name = project_name
+    end
+
+    def censored?(model, revisions)
```
corps, line 9: This is a super awkward interface.  It uses `?` notation to suggest it is a predicate, but it returns nil in all cases, and instead yields strings indicated reasons for censorship.

Would it be possible to change this to two functions:

```
def censored?
  censored_reasons.length > 0
end

def censored_reasons
 result = []
 return result unless restrict?
 result << *restricted_identifiers
 ...
 result
end
```

I feel this would still be easy to test using the `includes?` matchers and it would be a bit more obvious an interface.  You can even pump arrays into each other pretty simple ala `@errors << *censored_reasons`

corps: It would also make it possible to cache the result via an attribute `||=` operation, where as that is not possible with a block yield.

lib/magma/attribute.rb
graft: It occurs to me that this was a result of the previous version of Magma::Attribute, which was overloaded with a bunch of different type behaviors. These have been separated out into various actual attribute types (e.g. IntegerAttribute) which each define their own database_type - this method should ideally be broken up into those attribute classes (with the default here just `[ name, new_value ]`)

graft:
>I guess I'm not sure how to review.

There are many problems with the way this branch went down that I am not happy about which make reviewing this difficult - the breadth of this change (to using Magma::Loader instead of Magma::Model for /update) and its effect on the ability to post uploads ended up coupling this to the Magma-Metis integration.

Not much needs to happen here, I think this PR should be about: 1) as zach says, does it work? and 2) the work cole did building ontop of the Magma::Loader stuff.

This latter I think we should all take a look at, and make note of the problems around them, but they need not be solved here. Nor am I confident I know how these problems should be solved. In a nutshell, part of the conceptual work being done here is uncovering various document formats a Magma record transits through as it oozes through the system, including:

`Query` - the results from the Magma::Question apparatus
`Payload` - the main response formatting class, actually coming in json and tsv flavors.
`Revision` - the input format expected by the /update endpoint
`Loader` - a class now responsible for managing all manner of inserting of data into Magma
`Entry` - a format used by the Loader class to make SQL insertions

Each of these has varying behavior across each of the dozen-odd Magma attribute types (`string`, `integer`, `collection`, etc.).

All of this format-translation work had been going on for years, unrecognized and unlabeled, and consequently splayed out in various places in the codebase. Here, as a result of the somewhat panicked recognition that this stuff is going on, I have attempted to at least make this translation more explicit by renaming methods to reflect this x_to_y transition. Still, the result is decidedly obtuse and I consider the problem to be unsolved; let us consider it noted and move on, hoping to find a better abstraction in the future, amen.

coleshaw:
Hi all, thanks for the feedback. Yes, this is a tough PR to review because it's got so many things going on, but I certainly appreciate the comments, including all the style stuff.

I've addressed the comments. If y'all see other issues or have other feedback, we can create issues for those.

I'll let this percolate for a little bit more today, and if there is no additional feedback I'll merge and get it onto staging for further review. Will also be editing the Chef roles / configs to get staging to work.

PR #181, opened 2020-07-07
Add add_attribute action to update_model API

### LINKED ISSUE #169: Support add_attribute
In /update_model, we may add an attribute to an existing model using the `add_attribute` action. This is the simplest possible 'safe-migrating' action, which requires us only to add a new column.

We already insert attributes in a couple of places, for example in the load_projects command: https://github.com/mountetna/magma/blob/fb3154d877708d9f74fce257cf016b6ac08dd96f/lib/magma/commands.rb#L33
The requirements of the action are therefore the same as here, a valid input row include `model_name, attribute_name` and all of the `Magma::Attribute.options` (except maybe creation/update stuff).

The action should fail if the attribute exists - while we could update the existing attribute instead, the ambiguity is worth avoiding in such fraught situations.

In the non-migrating action we had a single job to do, to update the attributes table. This action, however, must also modify the data tables, in addition to inserting into the database and loading the new attribute. The migration here is pretty simple, just add a column to the corresponding model table with the specified attribute name.

How should we run the migration itself? Sequel provides a migration class which we use for running the migrations we normally use - we can (?) do the same here. In this case we might make a service class inheriting from Sequel::Migration which has an #up method that calls `alter_table` and `add_column` with the instance variables `@model_name, @attribute_name, @type`, which we can initialize with our arguments and then immediately apply. We don't need a down method because we should undo this action by running another action like `remove_attribute`, so we don't need a concept of reversibility here.

Once we've run the migration, the new column and attribute should be available to add data into through the usual Magma methods.

alimi:
Now that we're supporting two actions (`add_attribute` and `update_attribute`), we can see what behavior is shared among all actions. This has been encapsulated in `Magma::BaseAction`.

`add_attribute` is the first attribute that requires a migration. Instead of having every action perform migrations, we'll perform them once in `Magma::ModelUpdateActions`. After running migrations, `Magma::ModelUpdateActions` will restart the server to make sure Sequel has up to date info.

This PR also adds validations to `Magma::Attribute` via [Sequel validations](https://sequel.jeremyevans.net/rdoc/files/doc/validations_rdoc.html). The Sequel validations get used to validate the actions where possible.

Fixes #169


lib/magma/actions/add_attribute.rb
```diff
@@ -0,0 +1,104 @@
...
+  class AddAttributeAction < BaseAction
+    def perform
+      save_attribute
+      return false if @errors.any?
+
+      model.load_attributes([attribute])
+      true
+    end
+
+    def rollback
```
alimi, line 11: Actions can define `#rollback` for any changes that won't get rolled back from a failed database transaction. We mostly need this to clear out state that's saved in memory.

https://github.com/mountetna/magma/blob/718139d8c82cc1191da529f9bbf6226b45479127/lib/magma/actions/model_update_actions.rb#L13-L24

lib/magma/actions/model_update_actions.rb
```diff
@@ -1,26 +1,71 @@
...
     def perform
-      valid? && @actions.all?(&:perform)
+      return false unless valid?
+
+      Magma.instance.db.transaction do
+        raise FailedActionError unless @actions.all?(&:perform)
+        update_model_tables
+        true
+      end
+    rescue
```
alimi, line 21: I let this rescue `StandardError` because I'm not sure what errors we'll get if migrations fail to run. I couldn't figure how to run an action that would cause a migration to fail.

lib/magma/actions/model_update_actions.rb
```diff
@@ -1,26 +1,71 @@
...
+      @errors.map(&:to_h) + @actions.flat_map(&:errors)
     end
 
     private
 
+    def update_model_tables
+      migrations = @project.migrations
+      return if migrations.all?(&:empty?)
+
+      sequel_migration = eval("
```
alimi, line 36: Calling `eval` is a little risky, but we're not running user generated input so we should be OK. It feels like a decent trade off to get the migrations running. In the future, we might want to explore building the migrations without depending on strings.

lib/magma/actions/model_update_actions.rb
```diff
@@ -1,26 +1,71 @@
...
+        end
+      ")
+
+      sequel_migration.apply(Magma.instance.db, :up)
+      restart_server
+    end
+
+    def restart_server
+      return if Magma.instance.test?
+      Process.kill("USR2", Magma.instance.server_pid)
```
alimi, line 50: `USR2` is normal/hot restart. From the puma docs

>A hot restart means that no requests will be lost while deploying your new code, since the server socket is kept open between restarts.
>
> But beware, hot restart does not mean that the incoming requests won’t hang for multiple seconds while your new code has not fully deployed. If you need a zero downtime and zero hanging requests deploy, you must use phased restart.

https://github.com/puma/puma/blob/e2b419341153ff2206e1cb84fca1663639d0ef30/docs/restart.md

graft:
We tested this out in dev and it works! Our ability to model has expanded dramatically :)

Some issues I have found:

1) If you leave out project_name, the endpoint does not work; there should be a require_params somewhere.
2) If server_pid is not in config.yml, things break in strange ways; this is maybe okay.
3) Adding link_attributes would seem to be some kind of a farce because you cannot add both sides of a reciprocal link in a single operation. We'll probably have to handle this with a different operation.

alimi:
What error do you get when `project_name` is missing? `Magma::ModelUpdateActions#valid?` should return false when it's missing and the controller should return a 422.

```ruby
class UpdateModelController < Magma::Controller
  def action
    model_update_actions = Magma::ModelUpdateActions.build(@project_name, @params[:actions])

    if model_update_actions.perform
      # …
    else
      failure(422, errors: model_update_actions.errors)
    end
  end
end
```

https://github.com/mountetna/magma/blob/8dda698dab48a8d6205b3e19b19ee6d4f4fc06a3/lib/magma/actions/model_update_actions.rb#L13-L14

https://github.com/mountetna/magma/blob/8dda698dab48a8d6205b3e19b19ee6d4f4fc06a3/lib/magma/actions/model_update_actions.rb#L53-L65

graft:
This problem seems to originate in Magma::Controller, which seems to assume the existence of @params[:project_name] for each endpoint and provides @project_name, but does not require it. I don't think there are any tests catching this error, which is presumably shared across several endpoints.

```
WARN:2020-07-09T15:49:19+00:00 User saurabh.asthana@ucsf.edu calling update_model#action with params {:actions=>"[{:action_name=>\"add_attribute\", :type=>\"string\", :model_name=>\"project\", :attribute_name=>\"faq\", :description=>\"Frequently asked questions about COMET\", :di
splay_name=>\"FAQ\"}]"}
ERROR:2020-07-09T15:49:19+00:00 hxlkr Caught unspecified error
ERROR:2020-07-09T15:49:19+00:00 hxlkr undefined method `to_sym' for nil:NilClass
ERROR:2020-07-09T15:49:19+00:00 hxlkr /home/graft/magma/lib/magma.rb:31:in `get_project'
ERROR:2020-07-09T15:49:19+00:00 hxlkr /home/graft/magma/lib/magma/actions/model_update_actions.rb:68:in `initialize'
ERROR:2020-07-09T15:49:19+00:00 hxlkr /home/graft/magma/lib/magma/actions/model_update_actions.rb:10:in `new'
ERROR:2020-07-09T15:49:19+00:00 hxlkr /home/graft/magma/lib/magma/actions/model_update_actions.rb:10:in `build'
ERROR:2020-07-09T15:49:19+00:00 hxlkr /home/graft/magma/lib/magma/server/update_model.rb:6:in `action'
```

graft:
Something else we noted: there is no definition in the single-table-inheritance model_map for FloatAttribute. Is there a way to load the plugin after these attributes are defined so they can be collected automatically?

alimi:
I went ahead and manually added `Magma::FloatAttribute` to the model_map for now so we're not blocked. https://github.com/mountetna/magma/pull/181/commits/780d163cd2fdc79676365abd3f5c444f620c4b7a I also opened an issue to make it automatically configured, https://github.com/mountetna/magma/issues/183. I'm gonna spend an hour or so to see if I can get it to work.

I also opened https://github.com/mountetna/magma/issues/184 to check for `project_name` in all server requests.

PR #182, opened 2020-07-13
Support add_model in /update_model API

### LINKED ISSUE #170: Support add_model
The data graph starts off with a single model - we wish to be able to grow this graph by adding new models.

At first glance this is a simple proposition - we only need to specify the name of the new model, the migration will create the new table, and *finis*. However, Magma wishes the models in the graph to be connected, and moreover to be connected in a tree with reciprocal links. A new, empty model is an orphaned subgraph and cannot be allowed. This means that the new model must be added with a valid link to an existing model.

This makes the input requirements of the `add_model` action slightly more complex: we must specify a parent model that this model attaches to. Furthermore, we must specify the relationship: the parent may link back using the `child`, `collection` or `table` attributes. So let's say we require `model_name, parent_model_name, parent_link_type`. Since Magma prefers string record identifiers, we might also require they name the new model's `identifier`, unless `parent_link_type` is `table`.

Now we have several tasks to perform: we must insert a new row into the model table, two or three new  rows in the attributes table (one for the parent attribute in the new model, one for the link in the parent model, possibly one for the identifier in the new model). We must load the new model and attributes.
In our migration we must create a new table for the model and add the new columns for the parent and identifier attributes.

alimi:
This PR builds on top of https://github.com/mountetna/magma/pull/181.

`Magma::AddModelAction` will add the new model and define link relationships. It makes changes to both the database and in memory so we can build and run migrations for the new model in `Magma::ModelUpdateAction`.

This PR also makes a change to `Magma::BaseAction#validate`. Before, it would stop doing validations if any action validation failed. While this might save some code cycles, it could lead to a poor user experience because users would have to make multiple requests to find all validation issues. Now `Magma::BaseAction#validate` will let all validations run so we can collect all validation problems in one request cycle. Along the same lines, `Magma::UpdateAttributeAction#validate` has been updated to do more validation checks before calling `#perform`.

Additionally, `Magma::ModelUpdateActions#perform` has been updated so it reports any unexpected errors. It was capturing these errors, but it wasn't collecting them for reporting. This would lead to an empty `errors` object in the response payload that makes debugging difficult. Now, `Magma::ModelUpdateActions#perform` will collect these errors so they get reported back to users.

Fixes #170

### DIFF
--- a/lib/magma/actions/add_model.rb
+++ b/lib/magma/actions/add_model.rb
@@ -47,7 +47,7 @@ class Magma
 
     def validate_required_fields
       required_fields.each do |field|
-        validate_pressence(field)
+        validate_presence(field)
         validate_snake_case(field)
       end
     end
@@ -58,7 +58,7 @@ class Magma
       fields
     end
 
-    def validate_pressence(field)
+    def validate_presence(field)
       return if @action_params[field] && @action_params[field] != ""
 
       @errors << Magma::ActionError.new(
--- a/lib/magma/actions/add_model.rb
+++ b/lib/magma/actions/add_model.rb
@@ -1,0 +1,37 @@
+class Magma
+  class AddModelAction < BaseAction
+    def perform
+      Magma.instance.db[:models].insert(
+        project_name: @project_name,
+        model_name: @action_params[:model_name]
+      )
+
+      model = project.load_model(
+        project_name: @project_name,
+        model_name: @action_params[:model_name]
+      )
+
+      project.models[model.model_name] = model
+
+      if @action_params[:parent_link_type] != "table"
+        model.identifier(@action_params[:identifier].to_sym).save
+      end
+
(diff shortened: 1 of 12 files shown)


lib/magma/actions/add_model.rb
```diff
@@ -0,0 +1,114 @@
...
+      [
+        :validate_required_fields,
+        :validate_parent_model,
+        :validate_parent_link_type
+      ]
+    end
+
+    def validate_required_fields
+      required_fields.each do |field|
+        validate_pressence(field)
```
graft, line 50: This is the only issue i can find in this PR.

alimi: Spelling is hard 🙈. Fixed in https://github.com/mountetna/magma/pull/182/commits/dc23c64f66560027845a902c490e0eefe899c634.

PR #185, opened 2020-07-13
Use a Proc for Magma::Attribute single table inheritance model map

### LINKED ISSUE #183: Automatically configure Magma::Attribute single table inheritance plugin
The plugin is currently configured manually in a `model_map`:

https://github.com/mountetna/magma/blob/c3e5ff98a2150acca518d23237ac1ce432bb4390/lib/magma/attribute.rb#L1-L18

When attributes are added or removed, we also have to remember to add or remove them in this `model_map`. Nothing errors if it gets out of date, so it's easy to forget about.

It'd be nice if the plugin/`model_map` could be configured automatically.

http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/SingleTableInheritance.html

alimi:
The Sequel single table inheritance plugin allows for the model_map to be defined using a Proc. The Proc will receive the `type` column value and convert it to a `Magma::Attribute `class (e.g. `Magma::StringAttribute`).

Fixes #183

### DIFF
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -1,22 +1,7 @@
 class Magma
   class Attribute < Sequel::Model
-    plugin :single_table_inheritance, :type, model_map: {
-      "parent" => "Magma::ParentAttribute",
-      "string" => "Magma::StringAttribute",
-      "match" => "Magma::MatchAttribute",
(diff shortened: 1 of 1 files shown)


alimi:
I find something new everyday 😄

PR #186, opened 2020-07-15
Add test for TSV + collection; update collection query_to_tsv

coleshaw:
This PR separates the Collection `payload` format from its `tsv` format. Previously the `tsv` format called `query_to_payload` internally, which caused the collection value to attempt to reformat / flatten twice, resulting in exceptions (and would result in users not being able to download TSV files in Timur, if they involved a collection attribute). This change removes the second formatting attempt and also adds a test to check behaviour.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -89,7 +89,7 @@ GEM
       byebug (~> 11.0)
       pry (~> 0.10)
     public_suffix (4.0.3)
-    rack (2.2.2)
+    rack (2.2.3)
     rack-test (1.1.0)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -89,7 +89,7 @@ GEM
       byebug (~> 11.0)
       pry (~> 0.10)
     public_suffix (4.0.3)
-    rack (2.2.2)
+    rack (2.2.3)
```
coleshaw, line 92: This should be consistent with other apps, I think.

Issue #196, opened 2020-07-21
Consolidate Revision and RecordEntry classes


graft:
Recently we resolved #40, which changed the /update endpoint to use Magma::Loader. This created some rather difficult situations in the code, much coupling between classes and lack of clarity about who is doing what. Herein I will attempt to summarize some of these issues; I will briefly allude to the others.

The inputs to /update are wrapped in the `Magma::Revision` class, which previously would use `Sequel::Model.update` to handle making changes. Recently we replaced this with using the `Magma::Loader` class, which uses `Sequel::Model.insert` instead to do bulk insertions. The result is a data transit that looks something like this:
```
json => UpdateController => Revision => RecordEntry => entry hash => Sequel::Model.insert
```
Once the insertion is made, the results are reported out into a `Magma::Payload`:
```
Revision => Payload => json
```
However, this export is done via `Revision`, not `Loader`/`RecordEntry`.

Each of the transitions is managed by methods in the Attribute classes, which must transform each of the values in a record differently to accomplish each transition. The result is a lot of coupling: the controller, Revision, Loader and each of the attributes are all passed back and forth, and much ugliness ensues.

The main cause of this mess is the redundancy between the Revision class and the RecordEntry class. Crucially, the Revision wraps a single record (reflecting its use of the atomic `Sequel::Model.update`), while the Loader wraps the entire set of revisions, that is, a set of records for each model (reflecting its use of the bulk `Sequel::Model.insert`). The Loader, in fact, uses the RecordEntry class for wrapping individual records - this class is responsible for formulating the raw insertion hash used by `insert`. The `Revision`, therefore, must pass values into RecordEntry, which it does using the misnamed `Magma::Attribute#revision_to_loader`, while the RecordEntry formulates valid insertions using `Magma::Attribute#entry`.

To improve the situation, we may move whatever logic exists in `Revision` into `RecordEntry`, which can be instantiated directly in the `UpdateController`. The `Revision` and all of its associated `Attribute` methods can be removed - whatever careful formatting work remains can be done by `Magma::Attribute#entry`.

Since `RecordEntry` already receives the results of `Sequel::Model.insert`, the `Loader` may also now report directly out to the `Payload`, rather than having the more ignorant `Revision` attempt to do so.

PR #197, opened 2020-07-21
Support table attribute in /update

graft:
This adds some basic support in the /update endpoint for updating `table` attributes. There were two main issues to resolve:

1) It is customary to use `table` to link to a leaf model with no identifier (that is, just an ancillary table attached to this model; the only relevant identifier is the parent's), in which case many Magma operations substitute the database `id` column for the identifier. The /update revision format, however, posts changes as { identifier => { identifier_changes } }. If we are inserting new records, there must be a way for us to specify a non-existent id (that is, cause the insertion to create a new id).

2) There might be existing records in a table that should be removed before inserting new values.

Here I have implemented (1) and not (2). I think it is possible to do (2) but not without increasing the coupling across several Magma classes, which I am not willing to do, and will instead opt for the somewhat more painful route of deleting records in the console manually before re-inserting (if necessary) to meet our v1 goals.

(1) is mostly accomplished by resurrecting the old `TempId` system - this allowed a way to create temporary associations by wrapping objects (e.g. the hash representing a patient) in a `TempId`, which following the insertion into the database can be updated with the real id of the insertion, allowing us to resolve links before the /update completes (see `Magma::Loader#update_temp_ids`). Here I have reattached this system to the /update input using the special identifier `::temp`, followed by any unique string, e.g. `::temp1`, `::tempkevin`, etc., which may be used to generate a `TempId`.

Regarding said coupling, the main discovery I have made here (but not implemented) is how to resolve a good deal of the confusion currently at play in the code base. I don't think I can take the time to go through all of the changes required and meet our goals, so this discovery must play out later. Summarized in #196.

### DIFF
--- a/lib/magma/loader.rb
+++ b/lib/magma/loader.rb
@@ -46,7 +46,9 @@ class Magma
     end
 
     def push_record(model, record)
-      records(model) << RecordEntry.new(model, record, self)
+      id = record[:temp_id]&.is_a?(Magma::TempId) ? record[:temp_id] : record[model.identity]
+      records(model)[id] ||= RecordEntry.new(model, self)
+      records(model)[id] << record
     end
 
     # Once we have loaded up all the records we wish to insert/update (upsert)
@@ -91,7 +93,8 @@ class Magma
     def records(model)
       return @records[model] if @records[model]
 
-      @records[model] = []
+      @records[model] = {}
+
       ensure_link_models(model)
 
       @records[model]
@@ -108,7 +111,7 @@ class Magma
 
       @records.each do |model, record_set|
         next if record_set.empty?
-        complaints.concat(record_set.map(&:complaints))
+        complaints.concat(record_set.values.map(&:complaints))
       end
 
       complaints.flatten!
@@ -125,8 +128,8 @@ class Magma
         next if record_set.empty?
 
         # Our insert and update record groupings.
-        insert_records = record_set.select(&:valid_new_entry?)
-        update_records = record_set.select(&:valid_update_entry?)
+        insert_records = record_set.values.select(&:valid_new_entry?)
+        update_records = record_set.values.select(&:valid_update_entry?)
 
         # Run the record insertion.
         multi_insert(model, insert_records)
@@ -168,7 +171,7 @@ class Magma
     def update_temp_ids
       @records.each do |model, record_set|
         next if record_set.empty?
-        temp_records = record_set.select(&:valid_temp_update?)
+        temp_records = record_set.values.select(&:valid_temp_update?)
 
         MultiUpdate.new(model, temp_records.map(&:temp_entry), :real_id, :id).update
       end
--- a/lib/magma/loader.rb
+++ b/lib/magma/loader.rb
@@ -25,16 +25,6 @@ class Magma
 
   # A generic loader class.
   class Loader
-    class << self
-      def description desc=nil
-        @description ||= desc
-      end
-
-      def loader_name
-        name.snake_case.sub(/_loader$/,'')
-      end
-    end
-
     attr_reader :validator
 
     def initialize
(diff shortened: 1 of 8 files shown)


spec/update_spec.rb
```diff
@@ -593,237 +590,152 @@ def update(revisions, user_type=:editor)
...
-      }))
+      # the updated record is returned
+      expect(last_response.status).to eq(200)
+      #expect(json_document(:labor, 'The Golden Apples of the Hesperides')[:prize]).to match_array(Labors::Prize.select_map(:id))
     end
 
-    it 'removes an image reference using ::blank' do
-      lion = create(:monster, name: 'Nemean Lion', species: 'lion', selfie: '{"location": "https://metis.test/labors/Nemean Lion/headshot.png", "filename": "", "original_filename": ""}')
-
+    xit 'updates a table and returns the correct id' do
```
coleshaw, line 657: Is this test supposed to be skipped?

spec/update_spec.rb
```diff
@@ -593,237 +590,152 @@ def update(revisions, user_type=:editor)
...
-      }))
+      # the updated record is returned
+      expect(last_response.status).to eq(200)
+      expect(json_body[:models][:prize][:documents].values.map{|p|p[:name]}).to eq(['apple of joy']*2)
     end
 
-    it 'links an image from metis' do
-      Timecop.freeze(DateTime.new(500))
-      lion = create(:monster, name: 'Nemean Lion', species: 'lion')
+    xit 'replaces an existing table' do
```
coleshaw, line 708: Is this test supposed to be skipped?

graft: Yes, both of these should be skipped. I would like this behavior to work but didn't have time to implement it for now, so I opted to leave a skipped test in place as a reminder.

lib/magma/loader.rb
```diff
@@ -25,16 +25,6 @@ def entry value
 
   # A generic loader class.
   class Loader
-    class << self
-      def description desc=nil
```
coleshaw, line 29: Removing this method seems to break the IPI project. I get:

```
magma  | 	15: from /home/developer/magma/lib/magma/server.rb:13:in `initialize'
magma  | 	14: from /home/developer/magma/lib/magma/server.rb:13:in `tap'
magma  | 	13: from /home/developer/magma/lib/magma/server.rb:14:in `block in initialize'
magma  | 	12: from /home/developer/magma/lib/magma.rb:52:in `load_models'
magma  | 	11: from /home/developer/magma/lib/magma.rb:52:in `each'
magma  | 	10: from /home/developer/magma/lib/magma.rb:53:in `block in load_models'
magma  | 	 9: from /home/developer/magma/lib/magma.rb:53:in `new'
magma  | 	 8: from /home/developer/magma/lib/magma/project.rb:29:in `initialize'
magma  | 	 7: from /home/developer/magma/lib/magma/project.rb:84:in `load_project'
magma  | 	 6: from /home/developer/magma/lib/magma/project.rb:92:in `load_project_files'
magma  | 	 5: from /home/developer/magma/lib/magma/project.rb:92:in `require'
magma  | 	 4: from /home/developer/magma/projects/ipi/requirements.rb:18:in `<top (required)>'
magma  | 	 3: from /home/developer/magma/projects/ipi/requirements.rb:18:in `require_relative'
magma  | 	 2: from /home/developer/magma/projects/ipi/loaders/flow_jo.rb:3:in `<top (required)>'
magma  | 	 1: from /home/developer/magma/projects/ipi/loaders/flow_jo.rb:4:in `<module:Ipi>'
magma  | /home/developer/magma/projects/ipi/loaders/flow_jo.rb:5:in `<class:FlowJoLoader>': undefined method `description' for Ipi::FlowJoLoader:Class (NoMethodError)
```

Are you using a branch of the IPI project, or no longer using it? Or is there something I need to tweak locally?

coleshaw:
Outside of the issue with the `Loader` description method, looks good to me. Ran fine when I commented out the descriptions in each of the IPI loaders, but that should get addressed before deploying (either in Loader or in IPI), otherwise I worry the deployments will break.

PR #198, opened 2020-07-24
Configure Magma::Attribute sti key_map

alimi:
Without a `key_map`, the Sequel single table inheritance plugin will use the `model_map` to insert values into the database. For `Magma::Attribute`, this resulted in types being inserted as Ruby classes (i.e., `Magma::IntegerAttribute`) instead of strings (i.e., `integer`). By configuring `key_map`, we can make sure `Magma::Attribute` types are inserted as strings.

Info on the plugin can be found here: http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/SingleTableInheritance.html.

### DIFF
--- a/lib/magma/attribute.rb
+++ b/lib/magma/attribute.rb
@@ -1,7 +1,8 @@
 class Magma
   class Attribute < Sequel::Model
     plugin :single_table_inheritance, :type,
-      model_map: Proc.new { |type| "Magma::#{type.classify}Attribute" }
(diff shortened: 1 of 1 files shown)


alimi:
Also, if there's a bunch of records that need to be updated in prod you can run something like the following on the console.

```ruby
Magma::Attribute.each do |attribute|
  attribute.update(type: attribute.type.match("Magma::(.*)Attribute")[1].underscore)
end
```

You'll have to restart the server though so the cached attributes in memory get reloaded.

PR #199, opened 2020-07-24
Add column_name to the attributes table so attribute_name is editable

alimi:
With this change, a `Magma::Attribute`'s `attribute_name` can be changed at any time without also requiring a migration to the `Magma::Attribute`'s project table. This is accomplished by adding a `column_name` column to the `attributes` table.

To help verify this new behavior, I changed a few `attribute_name`s to make them different from their `column_name`s. A few things came out as a result:

1. Change `Magma::Model.identity` so it holds the identifier attribute instead of a symbol. Now that `attribute_name` and `column_name` can be different, a symbol with one of those values is no longer sufficient.
1.  `Magma::Model.identity` will fallback to a default primary key identifier for models who don't specify identifiers. This will make `Magma::Model.identity`'s behavior consistent across the board.
1. Pass full attribute object to `Magma::ColumnPredicate`s so the predicates can get `column_name` and `attribute_name` as needed.
1. Define alias methods on `Magma::Model`s for attributes that have different `attribute_name`/`column_name` because Sequel will only define methods based on `column_name`.


db/migrations/006_add_column_name_to_attributes.rb
```diff
@@ -0,0 +1,23 @@
+Sequel.migration do
+  up do
+    alter_table(:attributes) do
+      add_column :column_name, String
+    end
+
+    Magma.instance.db.execute("UPDATE attributes SET column_name=attribute_name")
+
+    alter_table(:attributes) do
+      set_column_not_null :column_name
```
alimi, line 10: We'll add the `column_name` column and backfill existing rows before making the column not nullable.

lib/magma/model.rb
```diff
@@ -22,6 +22,11 @@ def load_attributes(attributes = [])
           @parent = attribute.name if attribute.is_a?(Magma::ParentAttribute)
           attribute.magma_model = self
           self.attributes[attribute.name] = attribute
+
+          if !attribute.is_a?(Magma::Link) && attribute.attribute_name != attribute.column_name
```
alimi, line 26: We won't define aliases for link attributes with different `attribute_name`/`column_name` because they'll have Sequel association methods instead.

